Issue #24, opened 2024-10-16
[PCB] Improve VHF harmonic suppression


VanceVagell:
Improve the VHF filtering, I'd like the second harmonic to be even more supressed (currently it's just over -40dB, which is the minimum for part 97).

VanceVagell:
I already have a v1.6 design that is meant to fix this, should have test boards soon to confirm.

nhendin:
I've just ordered a v 1.6 PCB,  and have access to decent spectrum analyzers at work.  (Keysight portable as well as an Rohde & Schwarz FSV3 series). 

What's your test setup?  I assume it's  just transmit into an appropriate attenuator (or coupler to a dummy load) and measure 2nd harmonic suppression?  

I can attempt to duplicate the test that when I get my boards and parts set up (I just ordered a set of stuff from Amazon / Aliexpress / Signal Stuff) yesterday so it will be a little while.

VanceVagell:
Thanks for the offer, more testing is always welcome. I test with a TinySA spectrum analyzer, transmitting through a 30dB attenuator (the TinySA usually adds a few more dB of attenuation as it figures out the graph, something like 37dB total usually). I'm expecting the v1.6 PCBs to be delivered today and will update here after my own testing.

If you have any PCB design experience, I'd really love any help improving it overall. I suspect there are many RF-related improvements that could be made (e.g. better via stitching, impedence matched traces, etc.). I'm fairly new to PCB design.

nhendin:
What's the spec you are trying to hit for harmonic suppression?
If it's relative (like -40dBc) then the total attenuation doesn't matter as
long as the level is low enough to not cause non-linearities (e.g. more
harmonics) in the spectrum analyzer itself.
If it's an absolute spec, (e.g. less than -45 dBm) then calibration of the
attenuation might matter.

--Neil.

nhendin:
AFAIK, it looks like 47 CFR, Chapter I, subchapter  D, part 97, section 307(e) is the relevant document:

[...]

PR #26, opened 2024-10-16
Translated PCB to KiCAD

DefProc:
Translated the board into KiCAD 8. 

Open for double checking, but if you'd like to include this that's fine. 

PDFs for review:
[kv4p-ht-1.6-schematic.pdf](https://github.com/user-attachments/files/17401311/kv4p-ht-1.6-schematic.pdf)
[kv4p-ht-1.6-pcb.pdf](https://github.com/user-attachments/files/17401312/kv4p-ht-1.6-pcb.pdf)


kiavash-at-home:
It seems this is based on version 1.6 which is removed by the latest commit 0db745ceab97506fc33ee2fb4e05a766d37e18f9

DefProc:
Yes, this was translating the 1.6 board. I see from #24 that 1.6 does not have a suitable low-pass filter design. 

I'm updating this to match the 1.7 design, but I think I can see a better RF layout here.

VanceVagell:
Hi Patrick, thanks for working on this. There are now 1 or 2 other
electronics engineers in the #contributors channel on Discord also looking
at PCB updates. You might want to coordinate, or at least consider
different variations you each want to contribute.

kiavash-at-home:
I have looked at the latest KiCAD layout design in [29170bf](https://github.com/VanceVagell/kv4p-ht/pull/26/commits/29170bfd1de29698bacae1e9b3199fc97c5cc267). The filter layout is certainly much more improved than the one posted on EasyEDA. I only have one comment, filters in general are very sensitive to any mutual-coupling between their inductors, I highly recommend to rotate L2 by 90 degrees making it orthogonal to L1 and L3. L1 and L3 seem to be physically far enough that their magnetic field won't couple into each other.

VanceVagell:
It looks like this is still a WIP, please let me know when it's ready for another look (and attach a PDF again, that's helpful). I appreciate your hard work on improving the design and making it available in KiCAD format for people!

DefProc:
> I have looked at the latest KiCAD layout design in [29170bf](https://github.com/VanceVagell/kv4p-ht/pull/26/commits/29170bfd1de29698bacae1e9b3199fc97c5cc267). The filter layout is certainly much more improved than the one posted on EasyEDA. I only have one comment, filters in general are very sensitive to any mutual-coupling between their inductors, I highly recommend to rotate L2 by 90 degrees making it orthogonal to L1 and L3. L1 and L3 seem to be physically far enough that their magnetic field won't couple into each other.

I found a specific reference to this in some testing done [by W0QE](http://w0qe.com/Technical_Topics/coupling_between_coils.html) that seems to suggest the effect is minimal for ferrite cored - presumably because the core is discontinuous between the two inductors - but it's not something I've tested for. 

I agree that filter layout will be important going forwards - hence the current "build and test" cycle of investigation - and guidelines like this are really good for narrowing in on the best compromise of layout and efficiency.

Issue #20, opened 2024-10-16
[PCB] Apply RF best practices to the PCB design


VanceVagell:
Improve my existing PCB design, I'm very amateur at PCB layout. I think there may be an impedence mismatch going to the antenna and I don't think I have enough stitching with the ground plane (among other things!)

SmittyHalibut:
I provided some comments in https://github.com/VanceVagell/kv4p-ht/pull/50 on laying out the filter.

SmittyHalibut:
Done in #86.  Resolve please (it won't let me.)

PR #45, opened 2024-10-21
KiCAD PCB 1.5

DefProc:
Converted the 1.5 board to KiCAD. 

The intention here is to have an accurate translation of the existing known-good board (from EasyEDA) that can be used as the basis of futher development of the layout, or can be used as a reference for alternate components. This will allow sharing and variation of the source files, while keeping them within source control - along with the rest of the project. 

kv4p-ht-front
kv4p-ht-back

PDFs of the 1.5 translations: 
[kv4p-ht-pcb-1.5.pdf](https://github.com/user-attachments/files/17462050/kv4p-ht-pcb-1.5.pdf)
[kv4p-ht-schematic-1.5.pdf](https://github.com/user-attachments/files/17462051/kv4p-ht-schematic-1.5.pdf)


DefProc:
This replaces #26

DefProc:
Intended as a basis for #20 and for use in #24

VanceVagell:
Thanks for this! Hope it helps other contributors with making PCB updates.

PR #50, opened 2024-10-22
[PCB] Improved RF on 2-layer PCB layout in KiCAD [1.8]

DefProc:
Relating to #20 and #24

RF path has 1.06/0.2 trace/space for close to 50ohm impedance on the antenna waveguide. Via stitching is applied around the path, and board edges to bind the ground planes top and bottom. 

image
KiCAD calculator values

The low-pass filter design has in-line inductors (inductive coupling not checked) and capacitor pads in-line to minimise stub traces. 

image

Controller to Radio routing is moved mostly to the back side, with co-linear traces to maximise the ground plane continuity. 

image

I've swapped the generic footprints from easyeda defaults to KiCAD defaults. 

Still required: 
 - [ ] 3D model for the DRA818
 - [ ] 3D model for the ESP32 board
 - [ ] Full symbol fields detail for components
 - [ ] Verification of inductor and capacitor part selection (and footprint type)
 - [ ] Addition of 100nF and 47uF magic caps on the DA818 power pin (as per datasheet)

Possible extras: 
 - [ ] SMA Straight End Launch connector (below) - this would allow a lower profile case without the cutout for the SMA
 - [ ] combination footprint for right angle or board inline SMA connector

PDF files:

[kv4p-ht-schematic-1.8-draft1.pdf](https://github.com/user-attachments/files/17477576/kv4p-ht-schematic-1.8-draft1.pdf)
[kv4p-ht-pcb-1.8-draft1.pdf](https://github.com/user-attachments/files/17477577/kv4p-ht-pcb-1.8-draft1.pdf)


VegaDeftwing:
While it may be worth waiting for a 1.9, probably worth seeing if #54 should be squeezed into this change too, at least to testing it on some otherwise unused GPIOs

Bakermonitor1932:
Adding the pads for both smd and through hole addressable leds should be part of 1.7

SmittyHalibut:
Note that I haven't pulled down the code, applied this PR, and pulled it up in KiCAD.  This review is entirely based on the images and PDFs posted here.

* The board PDF is exported mirrored left to right.  Just cosmetic.
* ABSOLUTELY add decoupling caps on the 818 power pin.
  * Include a large bulk cap like a 47uF or even larger if you can fit it.  100nF isn't bad either.
  * Also include a tuned cap, something with a specifically low impedance at the frequency range in use.  Something like a Murata GCM1885C1H222 (JLCPCB: C882578), 0.04 ohms at 149MHz (see below)
  * When you get to making a UHF, Murata GCM1885C2A221JA16D (JLCPCB: C388907) 0.1 ohms at 436MHz (see below)
  * These will prevent any RF from the radios getting into the rest of the circuit, and will optimize decoupling their current needs.
* Turn the 818 module around 180 degrees.
  * This will allow the RF path to go directly into the filter, without having to pass parallel to it for so long.
  * It also brings your power pin closer to the source of its power, and gives you more space for decoupling capacitors.
  * It will mean moving the RF filter and connector to the other side of the board.  Is that feasible at this stage in development?
* Your filter inductors in a line like that will definitely have some coupling, which will negatively impact the filter performance.
  * These instructions are relative to the current layout. Adjust accordingly if you move the filter to the other side of the board.
  * Rotate L2 90 degrees so the radio side of it is closer to the board edge.
  * Flip C4 and C6 180 degrees, but keep them in roughly the same place.
  * Move L1 closer to the board edge.
* I'd also DC block the RF path.  Put a 1nF in series between the 818 and the filter (1nF * 50R = 3.18MHz).  This prevents any DC components from the radio.  It's PROBABLY DC blocked inside, but it's good to be safe.
* Vias don't cost money. Sprinkle those things all over the place.
  * You have vias at the corners of traces, which is very good.  That will help bridge ground over the traces.  Place more Vias along the lengths of the traces too.
* Put DC blocking caps on the audio traces; the DC bias points of the microcontroller and radio are almost certainly different.
  * MIC/MIC_OUT: Remove R1 (unless you know something I don't), put a 1uF in series close to the microcontroller, and a 10k resistor to ground right next to the 818.  (1uF * 10k = 16Hz cut off.)
    * You might also need a voltage divider here; the mic input is ~100mV full scale.  If that's what the 1M is doing, I suggest you don't rely on the internal impedance of the 818 and instead use a 10k/1k external voltage divider in place of the 10k load resistor right at the radio.
  * AF_OUT: same thing in reverse: 1uF in series close to the 818, 10k to ground close to the microcontroller.  Same 16Hz cut-off.
  * If you need a different cut-off point, f=1/(2*Pi*R*C)
* PD (Power Down) CANNOT BE PULLED HIGH.
  * (I don't think it'll blow anything up, but it'll totally break the biasing of the power amp inside the 818.)
  * If your software is only ever going active low, then high impedance, then this can work.  But I would feel WAY more comfortable if this was enforced in hardware.
  * I suggest an open drain MOSFET to ground, controlled by the microcontroller. AO3400A (JLCPCB: C20917, a basic part!) with a 10k from GPIO19 to the gate, source at ground, and drain on the PD pin on the 818.  Get it as close to the 818 as you can to minimize parasitics in the bias circuit.

|Z| of GCM1885C1H222 over frequency:
image

|Z| of GCM1885C2A221 over frequency:
image

SmittyHalibut:
As for the filter component value selection, this is how those values simulate in Elsie:

Schematic as entered:
image

Analysis: 100MHz to 1GHz, log scale.  Q = 10M (default, very not realistic)

Graph and markers:
image

## What the heck does all that mean?

-0.36dB at 148MHz is pretty darn good, but that's the best case of a filter.  Real parts are only likely to make this worse.  If possible, redesign with a knee a bit higher.

SmittyHalibut:
See #24 for my work on designing a new filter.

VanceVagell:
Looks like some great progress here, but I'll hold until the discussion and comments are resolved.

Brickerhaus:
Holy Cow!  You guys have been busy.  This is TekKnowology from discord by the way. 

- Where can I get the KiCad project file?  I have KiCad but I typically work in Xpedition, have experience with Cadence, Eagle, and Ultiboard.  I think I can get it to work without too much effort.
- The vias are interesting.  Vias are not free, but they are really really cheap.  Does KiCad allow you to put the stitching vias that are ringing the board in a more random pattern.  What is that current distance?  You generally want the distance to be random to the frequencies that you are encountering through the board.  But that is also a rule of "thumb" from an Signal Integrity expert.  The return vias look good too.  
- What is the decision on the caps at the power pin of the radio module?  
- On another wrinkle.  The wireless power delivery on a phone.  Does anyone know how that specifically activates?  Could we be unintentionally activating it and what effect would the traces on the other side of the board  nearest to the phone be affected by the wireless power delivery?

VanceVagell:
@DefProc committed a KiCad design version which you can find here: https://github.com/VanceVagell/kv4p-ht/tree/main/pcb/KiCAD/kv4p-ht

SmittyHalibut:
I hate seagulls. Stay with me here. Seagulls are people who fly in, make a bunch of noise, poop on your lunch, then fly away without adding any real value. I don’t want to be a seagull. Would it be ok if I applied the changes I suggested and submitted a pull request?  I’m not sure how to PR a PR though. If there’s a commit I should start from, let me know. On Oct 26, 2024, at 7:51 AM, VanceVagell ***@***.***> wrote:﻿
@DefProc committed a KiCad design version which you can find here: https://github.com/VanceVagell/kv4p-ht/tree/main/pcb/KiCAD/kv4p-ht

SmittyHalibut:
I've got all these changes made (and some more) in a commit that's coming.  I still need to add DC blocking caps to RF and Audio paths, and I did the 818/RF layout differently than I described above, but it accomplishes the same things.

Sneak peak:
image
image

Brickerhaus:
Wow.  What are the double pins for?

DefProc:
Cool,

Did you have to increase the board width to accommodate the rotated radio
module?

SmittyHalibut:
> Wow. What are the double pins for?

https://github.com/VanceVagell/kv4p-ht/issues/59  Multiple different ESP32 modules.

SmittyHalibut:
> Cool, Did you have to increase the board width to accommodate the rotated radio module?

A bit, yes.  But not a lot.

Issue #58, opened 2024-10-26
Case not 3D Printable by JLC3DP


m1geo:
As I was having a batch of PCBs made for my radio club, I figured I'd get JLCPCB 3DP to SLA print nice cases. However, they are unable to print the cases.

I know this isn't the intended use case, but thought others may try and so it may be worth collating their feedback. 

Below is the JLCPCB 3DP email reply:

[QUOTE]

Hello,

Hope this email finds you well.

Sorry to bother you, but there are some details that we need to confirm with you about your 3D printing order before proceeding.

1, As shown below, several separate shells are detected in your file, sorry to tell you that each file can only support 1 shell (1 model) print.

The different shells are shown in Green color. Please kindly unify all the separate shells into one shell.

Sorry for the inconvenience, but please advice, shall we activate the replace file button or you can cancel it?

![Screenshot_20241026_033143_BlueMail](https://github.com/user-attachments/assets/3d0da767-3663-4750-9a57-413ed6d90b4a)

2, The wall thickness of the indicated red area is not enough which has a high risk of deformation, crack and loss, could you please kindly increase it to be at least 0.8mm to proceed? 1.5mm will be much better for the final print.

![Screenshot_20241026_033159_BlueMail](https://github.com/user-attachments/assets/5edb2b30-3291-462a-a672-df238503813c)

3, The depth of the texts are not enough, which could not be fully printed out and the texts could be very blurry.

Could you please kindly increase the depth of the texts to be more than 0.8mm to get a better print?

![Screenshot_20241026_033203_BlueMail](https://github.com/user-attachments/assets/955e8ceb-2a8d-4c6d-b038-aee33061e9da)

Your early reply will be highly appreciated !

Best Regards

JLC3DP Team

[END QUOTE]

Cheers for a cool project!
George M1GEO.

Magnum164:
I volunteered with ASME on the 3D printing standard a few years back and a
lot of these comments are not uncommon.

For #1 the shells, they appear to be drawn in the CAD software as separate
objects and should be extruded and joined together. generally that is not
an issue but it is a reject for most certified 3D printing mfg sites. It
does indicate there is a gap between the solids.

#2 Really is picky and impacts design intent. With a .4mm nozzle .2mm is
not uncommon for a single slice. Not sure what the thickness is as I had my
friend print the case and can't find where I hid it right now. Their
concern may be it curling on them during printing and the cost of remakes,
which again goes to the machines they have, not the design and they should
[...]

PR #65, opened 2024-10-28
New Thicker base

Magnum164:
Increased the base by .408 and depth of text and exported as single STL to avoid potential rejection from JCLP and other print farms.
Also re-oriented the model to sit flat for printing without having to rotate in slicer.


VanceVagell:
Thanks for the improvements!

VanceVagell:
This should fix issue #58

PR #129, opened 2024-12-05
[PCB] v2.0c

SmittyHalibut:
v2.0c hardware.  Changes from v2.0b:
* Improved VHF LPF (lowered IL to 0.5dB, increased attenuation to >50dB)
  * Verified with LPF test boards.
* Added scallops to board edge for mounting.
* Added Vbus/2 sense to GPIO15 (measure Vbus, check for sagging)
* Added fuses to GPIO and SAO headers' power pins.
* Fixed I2C (added pull-ups that were missed before.)
* Added board revision detection. GPIO36, GPIO39. v2.0c is HIGH HIGH.
* Panelized 2 radios on one board, to meet 70x70mm minimum panel size for JLCPCB.

image image

image


jh4vsm:
Dear Sir,
I have a question about C10 and R1 for I034.
I think this capacitor is unnecessary for DC biasing from the DRA818 module to the ADC of the ESP32.
Alternatively, I think an additional pull-up resistor is needed along with the capacitor to bias I034 to the proper voltage (1.1V?).
Please forgive me if I am mistaken.
Thank you in advance.

SmittyHalibut:
> Dear Sir, I have a question about C10 and R1 for I034. I think this capacitor is unnecessary for DC biasing from the DRA818 module to the ADC of the ESP32. Alternatively, I think an additional pull-up resistor is needed along with the capacitor to bias I034 to the proper voltage (1.1V?). Please forgive me if I am mistaken. Thank you in advance.

C10 is required to break the DC between the ESP32 and the SA818, whatever that DC might be.  It's safe to include, even if its not needed.  It's cheap insurance.

You may be right about needing another resistor to bias the input to Vref/2.  The test boards we have are built the way shown (pulling down to 0v) and seem to work fine.  However, I'm asking @VanceVagell to record me some receive audio coming from the ESP32, to see whether we are getting clipping on the bottom of the waveform.

I think you're right though, the ADC isn't performing any internal biasing, and R1 is pulling it to GND, which should make half the waveform unreadable.

I've put a hold on my manufacturing order to resolve this.  Thank you for bring this up, @jh4vsm .  :-)

SmittyHalibut:
I added a 47k to +3.3v to the same line.  (Pseudomath follows.)

47k/10k * 3.3v = 0.58v
Vref = 1.1v (ish)

So a 47k/10k voltage divider should put the analog audio right in the middle of the ADC's range.

Issue #88, opened 2024-11-07
[PCB] Test v2.0b and improve v2.0c boards


SmittyHalibut:
Receive prototypes from #63 and test.  Find any changes needed.

Also: Add any additional feature requests here:
* **DONE** Current limiting/fusing on GPIO and SAO headers power pins
* **DONE** Pull-ups on I2C bus.  Oops.
* **DONE** Add board revision detection. I'm REALLY mad I didn't get that in v2.0.  Bleh.
* **VERIFYING** Raise cutoff frequency of LPF, lower insertion loss.
* **DONE** More sturdy PTT buttons
* **DONE** Add M3 scallops on the board edge.
* **DONE** Add a voltage detector to the VBUS from USB.
  * It's on GPIO15, with a 100k/100k voltage divider. So 5v will stay in range, and the pin can be used for other things instead if you don't care about measuring voltage.
* ADD MORE HERE (or, if you can't edit, just post to this ticket.)

SmittyHalibut:
Received boards from JLCPCB!  On to testing.

adoeller:
> On to testing.

did it explode? could you fill us in on the progress?

SmittyHalibut:
I'm keeping a list of things that need changing on the next revision in the top post of this thread.  So far, it's just fusing/current limiting some things, and adding pull-ups on the I2C bus on the SAO.  Nothing that impacts the functionality of the radio.

SmittyHalibut:
* Added 1.2k pull-ups on I2C
* Added 100mA resettable fuses to +5v on headers.
* Added 10k pull-downs on Sensor_VP and _VN, GPIO36 and GPIO39 for version detection pins.

SmittyHalibut:
Last thing I want to do is look at the LPF.

too long, didn't document: I want to switch from a Cheb to Cauer filter.  This involves adding capacitors in parallel with the series inductors.  Not a huge change, but it is a change to the board.  So now's the time to do it.  It's also a backwards compatible change; if you want to stick with the current design, just leave the caps unpopulated.

Using SVC Filter Designer https://tonnesoftware.com/svcfilter.html, I went with these design parameters:
* Order: 7
* Inductor Q: 30   (doesn't impact the design, only how realistic the simulation is.)
* Bandwidth (ripple): 208MHz   (aka, the cut off frequency.  See below for why this frequency exactly.)
* Impedance: 50
* Topology: C-input lowpass
* Family: Cauer
[...]

PR #134, opened 2024-12-11
Forgot to update LCSC part number for PTT buttons. #88

SmittyHalibut:
Updated the LCSC part number on the PTT buttons.  I had to change these by hand on JLCPCB's web page, so this applies that change permanently.


SmittyHalibut:
See #88

Issue #148, opened 2024-12-21
[PCB] Connecting Audio to ADC is Problematic


dkaukov:
### Issue
The current approach of directly connecting an AC audio source (±1V) to the ESP32 ADC input can lead to:

- Clipping: Audio signals may exceed the ADC’s 0-1.1V range, causing distortion.
- Biasing Issues: Without proper DC bias, the ADC cannot correctly sample the audio waveform.
- Noise Sensitivity: Floating or improperly biased inputs can pick up noise, leading to inaccurate readings.

### Suggestion
A simple way to improve this is to use GPIO26 (DAC2) to output the internal 1.1V reference. This can be used to bias the audio input, ensuring the signal stays within the ADC's measurable range.

```
  dac_output_enable(DAC_CHAN_1);  // GPIO26 (DAC1)
  dac_output_voltage(DAC_CHAN_1, 85 / 2);  // ~1.1V / 2 
```
When we can feed this voltage to ADC_IN (GPIO34) through 10k resistor.

This method leverages the ESP32's built-in Vref, reducing the need for external components while ensuring more accurate audio sampling.

VanceVagell:
Thanks Dmitry, this is all new information to me and very interesting. I
will have to bodge this in and test it out.

dkaukov:
This is what we have currently:

![IMG_20241222_103511353.jpg](https://github.com/user-attachments/assets/0bef3ff4-8b06-4ab4-bcbe-8a3330d4abdd)

PR #149, opened 2024-12-22
More complex approach to the audio "clicks"

dkaukov:
1) ADC bias injection (https://github.com/VanceVagell/kv4p-ht/issues/148)
2) ADC_ATTEN_DB_12 to be in spec
3) ADC dc removal
4) AudioFormat.ENCODING_PCM_16BIT as some Android devices do not support 8bit PCM (https://developer.android.com/reference/android/media/AudioFormat#ENCODING_PCM_8BIT)

All of this reduces chances of buffer underruns and when it happens, the click is almost not noticeable.


VanceVagell:
Thanks for the submission Dmitry. Does this fix the clicks for you?

dkaukov:
Yep, no more clicking for me.

dkaukov:
ADC Bias injection:
```
    DRA818 AOUT-+---- GPIO34 (ADC Input)
                |
               [R1] 22kΩ
                |
 GPIO26 (DAC2) -+
                |
               [C1] 0.1µF
                |
               GND
```
Discussed  https://github.com/VanceVagell/kv4p-ht/issues/148

VanceVagell:
Thanks for the hard work on this! Your example helped me identify the root cause, which was purely USB related. This is now fixed in commit cc1ee2319a45e9ba7151ffd514cdefc64e5914ee. Declining this PR, but it was instrumental to fixing this. Thank you!

VanceVagell:
If there are elements of this PR that you think are still useful, please break them up into smaller PRs that have limited functionality so I can review them each. Thanks again.

dkaukov:
Hi @VanceVagell ,

Thanks for the update and for fixing the USB-related issue! I appreciate the recognition of this PR. However, after testing commit [cc1ee23](https://github.com/VanceVagell/kv4p-ht/commit/cc1ee2319a45e9ba7151ffd514cdefc64e5914ee), the clicking issue still persists on my end. While reducing the sample rate lowers the click frequency, the issue hasn’t been fully resolved.

Initially, high bit rate serial transfer seemed like the primary suspect, so I conducted a series of error rate tests but didn’t detect any problems in my environment.

Regarding this PR, I believe it’s already atomic and relatively simple. If you’d like to test specific parts of the PR or need further details, feel free to let me know. I’d be happy to help with additional testing or adjustments.

Thanks again for your time and the work on this project!

VanceVagell:
If it's easy to do so, could you please attach an audio recording of the
remaining clicking sound you have with v1.5.0? I'd like to compare it to
other audio issues people have reported or I've heard with my test devices.
If GitHub won't let you attach it to a PR, please open an issue and attach
it.

I'm out of the house at the moment so I can't take a close look at the PR
to suggest how to make it more atomic, but from memory it had both USB
serial configuration changes AND new filtering logic.

Also, an initial test of the PR as-is did not work at 44.1kHz audio on my
test devices (audio not at all understandable) but somewhat worked at lower
sampling rates, but then APRS decoding no longer worked (does it work for
you with this PR)?

dkaukov:
Hi @VanceVagell ,

Thanks for the detailed feedback. I think the issue you're seeing with 44.1kHz might be due to testing the new firmware with the older Android app. Since the PR changes the wire audio format to ENCODING_PCM_16BIT, the existing app won't handle the audio correctly without corresponding updates.

Regarding APRS – you're right, it’s broken with the current PR because of the format change. The fix is straightforward: we just need to replace `convertPCM8ToFloatArray` with `convertPCM16ToFloatArray` to handle the new 16-bit format. I can make that adjustment if you'd like.

Also, here's the audio recording you requested (v1.5.0): https://www.youtube.com/shorts/QL5MdgGOt2E.

Thanks again for taking a look at this!

dkaukov:
Hi @VanceVagell,
Since the proposed changes depend on each other, I created "progressive" PRs in my fork:

[PR 1 – ADC Bias Injection](https://github.com/dkaukov/kv4p-ht/pull/1)
[PR 2 – ADC Attenuation Adjustment](https://github.com/dkaukov/kv4p-ht/pull/2)
[PR 3 – DC Removal](https://github.com/dkaukov/kv4p-ht/pull/3)
[PR 4 – Audio Encoding Format Change](https://github.com/dkaukov/kv4p-ht/pull/4)

Let me know if you'd like to review or discuss further.

VanceVagell:
Thanks so much Dmitry, this will make it easier for me to take a look, and test each part.

In the meantime, could you please test the latest build (v1.6.1 of app and v5 of firmware), and confirm if you still have the audio problems?

dkaukov:
Hi @VanceVagell,  

Thanks for the quick response! Unfortunately, I can't test the released v1.6.1 as I only have UHF modules, and I need to patch both the firmware and the Android app. After building from git, I can confirm the following:  
1. The clicks are much less frequent but still occur occasionally.  
2. The audio on RX seems distorted to me.  
3. The latency is noticeably high.  

These are the reasons behind the proposed patches. Let me know if you need more details or if I can assist further with testing.

VanceVagell:
Reopening now that this is split out. Will test.

Dmitry, does this still require the PCB modifications you previously outlined? If so, @SmittyHalibut should take a look at that to confirm if your change is compatible with the new PCB 2.x designs.

VanceVagell:
I tested this PR on both a v1.x PCB, and a v2.x PCB, with both the new Android app in this PR and the new firmware installed on the test device.

Unfortunately, I'm getting severe audio distortion on both PCB versions. I have not made the hardware modifications you mentioned in the other PR, would that cause these audio issues? We can't expect everyone to change their PCB, so any software change needs to work with the stock kv4p HT hardware.

Please let me know what you think is going on so we can move this PR forward. I'd love to merge it if it helps some people with other audio issues, but currently it's a regression for my test devices.

https://github.com/user-attachments/assets/c5d8cb8d-67a9-4a1c-8e4b-cc9fa9f93340
https://github.com/user-attachments/assets/ea900c2c-7f53-49b7-bac8-40fb4f942b1d

dkaukov:
Hi @VanceVagell,

Thanks for testing on both PCB versions and sharing your feedback. I completely understand your concern – software updates should work with stock kv4p HT hardware whenever possible.

That said, we can’t avoid physics – feeding an AC signal directly into the ESP32 ADC isn’t ideal, and staying within spec would require schematic changes. Apologies for that.

It looks like the samples you tested may have come from an incorrect merge. Could you try the pre-built binaries here to help narrow things down?
https://github.com/dkaukov/kv4p-ht/releases/tag/click-fix-01

Let me know how it goes. I appreciate your time!

dkaukov:
Hi @VanceVagell,

In my opinion, it’s better to keep this PR closed, as the code changes here are no longer mergeable (I didn’t expect it to be re-opened). Let’s focus on new discussions, starting with:
https://github.com/VanceVagell/kv4p-ht/pull/159

Thanks for your understanding!

dkaukov:
Code changes here are no longer mergeable

PR #153, opened 2024-12-26
fix function tuneTo, add argument,  string bandwitdth

takurx:
fix function tuneTo, add argument, string bandwitdth
I meet bug when build it on PlatformIO


VanceVagell:
Thanks for the fix!

PR #159, opened 2025-01-03
ADC bias injection

dkaukov:
- removes clipping
- makes louder

To inject bias you need hardware modification:
```
    DRA818 AOUT-+---- GPIO34 (ADC Input)
                |
               [R1] 22kΩ
                |
 GPIO26 (DAC2) -+
                |
               [C1] 0.1µF
                |
               GND
```

Fixes: https://github.com/VanceVagell/kv4p-ht/issues/148

### DIFF
--- a/microcontroller-src/kv4p_ht_esp32_wroom_32/kv4p_ht_esp32_wroom_32.ino
+++ b/microcontroller-src/kv4p_ht_esp32_wroom_32/kv4p_ht_esp32_wroom_32.ino
@@ -22,6 +22,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #include <DRA818.h>
 #include <driver/adc.h>
 #include <driver/i2s.h>
+#include <driver/dac.h>
 #include <esp_task_wdt.h>
 
 const byte FIRMWARE_VER[8] = {'0', '0', '0', '0', '0', '0', '0', '5'}; // Should be 8 characters representing a zero-padded version, like 00000001.
@@ -168,7 +169,7 @@ void initI2SRx() {
 
   // Initialize ADC
   adc1_config_width(ADC_WIDTH_BIT_12);
-  adc1_config_channel_atten(ADC1_CHANNEL_6, ADC_ATTEN_DB_0);
+  adc1_config_channel_atten(ADC1_CHANNEL_6, ADC_ATTEN_DB_12);
 
   static const i2s_config_t i2sRxConfig = {
       .mode = (i2s_mode_t) (I2S_MODE_MASTER | I2S_MODE_RX | I2S_MODE_ADC_BUILT_IN),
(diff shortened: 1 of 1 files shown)


SmittyHalibut:
What voltage does 132 on the DAC generate?

V2 hardware (and v1.9 when I get time to lay it out) already biases to 0.59vDC using a simple resistor divider on the regulated 3.3v supply. Vref is 1.1v, so 0.59v is just about the center. Meaning, v2 hardware should work well without any modifications.

microcontroller-src/kv4p_ht_esp32_wroom_32/kv4p_ht_esp32_wroom_32.ino
```diff
@@ -186,6 +187,8 @@ void initI2SRx() {
 
   ESP_ERROR_CHECK(i2s_driver_install(I2S_NUM_0, &i2sRxConfig, 0, NULL));
   ESP_ERROR_CHECK(i2s_set_adc_mode(I2S_ADC_UNIT, I2S_ADC_CHANNEL));
+  dac_output_enable(DAC_CHANNEL_2);  // GPIO26 (DAC1)
+  dac_output_voltage(DAC_CHANNEL_2, 138);
```
SmittyHalibut, line 191: What DC voltage does this generate?

dkaukov: Hi @SmittyHalibut, 

The SA818 datasheet lists the audio output amplitude at **0.7v**, though in practice it's closer to **1v Vpp**.  

To accommodate this, I opted for **ADC_ATTEN_DB_12**, which provides an input range of **150 mV to 2450 mV** ([Espressif ADC Reference](https://docs.espressif.com/projects/esp-idf/en/release-v4.4/esp32/api-reference/peripherals/adc.html)).  

Theoretically, the ideal bias point would be at the center of this range:   
\[(2450 + 150) / 2 = 1.3v\]  

However, in practice, I’ve found **1.75v** to work better for my boards.

dkaukov:
@SmittyHalibut This is my measurements after bias injection:

jh4vsm:
I think the audio output amplitude of 2V Vpp (as suggested by dkaukov) is true when the volume setting is 8.
For 0.59V DC bias and ADC_ATTEN_DB_0, a volume setting of 6 may be appropriate.
The second harmonic signal was generated at volume setting 8 and disappeared at volume setting 6 when a 1000 Hz monotone modulation was applied.

dkaukov:
Hi @jh4vsm, 

Reducing the volume with `ADC_ATTEN_DB_0` is a viable option, but its narrower input range  of ADC makes it more susceptible to noise, in my opinion.

VanceVagell:
Merging, as this does not affect audio quality on existing devices, and if it helps on some other devices that's great. Thanks for the PR.

dkaukov:
Hi @VanceVagell It seems like it was not merged..

VanceVagell:
Hmm I'm not sure what happened, it looks merged now. Thanks for letting me know!

PR #168, opened 2025-01-09
ADC auto DC removal

dkaukov:
- Handles ESP32 ADC reference voltage variations.
- Supports 16-bit I2S transfers.

This is a continuation of [PR #159](https://github.com/VanceVagell/kv4p-ht/pull/159). Here, we aim to automatically manage differences in the ADC Vref generator across chips.


VanceVagell:
Could you please merge the latest code from main into this? Then I'll test this one out.

I'm assuming that from a testing standpoint this change should be invisible to the end user, right? Audio etc. should all sound the same. If there's something more specific I should be checking for please let me know.

dkaukov:
Hi ,

I've merged the latest code from the main branch into this branch. From a testing standpoint, this change should be invisible to the end user—audio and other behaviors should remain the same.

VanceVagell:
Interesting re: revisiting the attenuation for the clicking, I will look into if that's still needed. It may be a few days before I can properly review and test this as I'm back to work. Thanks for sharing your improvements with the community!

VanceVagell:
Thanks. Not yet, I'm hoping to have time this weekend for it. Sorry for the
delay.

VanceVagell:
No problems found, merging. I started removing the special attenuation for squelch to see if that's needed, but it was taking too long to untangle how to get the audio out at the right volume. 

If possible, could you please test if that's needed any longer, and if not submit a PR to remove it?

Thanks for the PR!

Issue #135, opened 2024-12-12
[PCB] Document the order process for v2


SmittyHalibut:
After #88, the process for ordering boards has changed somewhat.  Write an ORDERING-BOARDS.md or similar.

roberthadow:
In more general terms--and thinking of the future--KV4P.com should allow the user to view and select from a list of supported boards. Subsidiary pages then would include features, requirements, and ordering instructions.

At the present time, four boards are of interest to the Discord group. This list will grow.
* 1.6 (as I remember). Not supported. EOL
* 1.7b  1.44 MHz only. Acceptable emissions.
*  1.7b.2 144 MHz only. Same board, different passive components.
* 2.0x Whatever stuff iit does that's special.

roberthadow:
It really is time to adopt and stick to a version numbering scheme.  There are two classes of interested people: the users and developers.  Users deserve an easy-to-understand numbering scheme.  They are concerned only with designs that have been tested and are ready for production.  Users are interested in four things:

**Mechanical**:  How big is the board and where are all its ports?  This determines what case they need.  Take for example v1.7b and 1.7b.2.  They are mechanically identical.  Their filters have different components. They take the same case.

**PCB Layout**: For people who want to buy bare boards and discrete components, they need to know if the board itself is the right one.  For example, v1.7b and 1.7b.2 have the same layout.  Choose different passive components and you have a different radio.

**Radio version** A radio is a PCB with all compoents specified.  From a user's perspective, it is a fully tested design.  At their option, they can order (1) a bare board and all components and build it themselves, (b) a board with most components affixed from China, or (3) a board with most components affixed from @SmittyHalibut.  

**Software version** For now, there are only two versions of software: (1) the one I downloaded from the Play Store and flashed my radio and (2) the one the Play Store has today.  Eventually, the app will allow a user to examine what's currently flashed into his or her radio, but that's on the the to-do list.  Serious hackers can use GitHub to reconstruct previous versions, but they are not in the regular user group. 

I suggest this hardware versioning scheme for public consumption. A radio is defined by a three-portion code.

[...]

PR #169, opened 2025-01-10
Quick README for how to order a v2.0 board. #135

SmittyHalibut:
Someone asked me about how to do this in Discord, and I wrote up some quick notes, and realized that those notes were basically the instructions I needed to write, once I added a bit of better formatting.  So here we are.

### DIFF
--- a/pcb/KiCAD/README.md
+++ b/pcb/KiCAD/README.md
@@ -1,0 +1,96 @@
+# Ordering kv4p-HT v2 boards from JLCPCB
+
+## Things you'll need:
+
+Use Chrome.  For some reason, the JLCPCB order process is broken for me on FireFox.  Much sadness.
+
+You'll need the following three files:
+* <https://github.com/VanceVagell/kv4p-ht/blob/main/pcb/KiCAD/kv4p-ht/production/kv4p_HT_2.0c.zip>
+* <https://github.com/VanceVagell/kv4p-ht/blob/main/pcb/KiCAD/kv4p-ht/production/bom.csv>
+* <https://github.com/VanceVagell/kv4p-ht/blob/main/pcb/KiCAD/kv4p-ht/production/positions.csv>
+
+## Order the PCB
+
+* Go to <https://jlcpcb.com/>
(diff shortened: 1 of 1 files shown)


SmittyHalibut:
R23 and R22 are just indicators for humans looking at the board.  Electrically, they're inert, connected to ground on both sides.  Think of them as a "Check mark" in a box next to each "VHF" and "UHF" text.

Whether a board is for VHF or UHF is entirely determined by the other parts: FL1, and all the Ls, Cs, and 0&Omega; Rs in the filter section.  And, ya know, which SA818 is soldered in place.  Not something that can be easily switched.  :-)

Pull Request:  I was talking about the documentation.  I'll publish the `production/` files for each band.  Not that I don't trust your files, it's just that it's probably easier than loading your files up and reviewing them.  :-)

n-connect:
> R23 and R22 are just indicators for humans looking at the board. Electrically, they're inert, connected to ground on both sides. Think of them as a "Check mark" in a box next to each "VHF" and "UHF" text.
> 
> Whether a board is for VHF or UHF is entirely determined by the other parts: FL1, and all the Ls, Cs, and 0Ω Rs in the filter section. 

Yepp, got you, hence my edit above - took some time to gather the infos around.

>And, ya know, which SA818 is soldered in place. Not something that can be easily switched. :-)

Questions here, from the three SA818S JLCPCB offers, the -V -U tags speaks for itself of course. I've read the [specs of the SA818S-CE](https://www.nicerf.com/walkie-talkie-module/walkie-talkie-module-sa818s-ce.html), and it shows both the freq range of UHF and VHF both -> so I guess misunderstood as it can do both. In your opinion, what will the SA818S-CE be, UHF or VHF? That's the only one JLCPCB let you to assemble on Economic PCB, so kind of a key question for me, who solders badly :)
In general is there such an SA818S-* which can do both? So a 2-band radio can be built from it? 
 
> Pull Request: I was talking about the documentation. I'll publish the `production/` files for each band. Not that I don't trust your files, it's just that it's probably easier than loading your files up and reviewing them. :-)
Wise choice :) , although every PR should be crosschecked by design. At least this I would do. I'm looking forward to see those updates incorporated. I like, but not dependent on the metal case, which is one of the main points 2.1b marked with. Also having 3D printer(s) around. In your view what other aspects 2.1b is a better choice over 2.0[d..c]? I wanna avoid falling for the versioning, after my misunderstanding of the 2-band capability (comes from 2/3-bands DMRs around) :)

Issue #166, opened 2025-01-09
[Android, ESP32] after at least 1 tx, transmit may stop working


VanceVagell:
I believe this is a software/firmware issue, not RFI.  This is on kv4p HTs that previously worked fine after tx, so it's not an RFI issue.

The app looks like its transmiting (on the UI) but it's not actually keying up the radio.

This suggests that the ESP32 is not receiving or perhaps not properly handling the PTT_DOWN command and going into MODE_TX.

This is a fairly high priority bug because transmit is obviously a core feature.

dafyddwalters:
I've also noticed something similar. To be fair I only noticed this since v1.6.2.

When I have squelch enabled (not set to 0 level) I can not make two tranmissions in quick succession. In other words: Tx for 1 second. Wait a second. Try to tx again. No tx.  Wait several more seconds, tx eventually works again.

More worryingly, while doing this test a few times, I found that the sometimes the tx carrier spontaneously started (no modulation/voice) after several seconds, and didn't stop until I pulled the USB cable (pressing the PTT again made no difference once it had gotten into this state).

The first issue (being unable to transmit for a few seconds) is easily repeatable. 

The second one (spontaneously going into tx and locking there) is not so easy to reproduce. I've seen it twice now.

As a sanity check I reverted back to commit 6400a1a, rebuilt and deployed the app but the firmware remains updated to version 6, and the "cannot make two transmissions in quick succession" issue remains. 

As I only started seeing the issue since v1.6.2 I wondered if perhaps the problem in the ESP32 changes for firmware version 6, but that could be a red herring.

VanceVagell:
This line is the most suspicious starting point to investigate:

https://github.com/VanceVagell/kv4p-ht/blame/d803f0940618c6a935aae182d4083e0b74cf1cd0/microcontroller-src/kv4p_ht_esp32_wroom_32/kv4p_ht_esp32_wroom_32.ino#L146

dafyddwalters:
Relating to the second part of the issue I desdibed earlier, I should also probably mention that I almost always use the PTT in "toggle" mode rather than "hold down" mode, as I suppose perhaps this may have some bearing on the issue where the kv4p HT starts transmitting spontaneously. 

Just mentioning it in case it's significant.

It's almost as if when it fails to transmit the second time, a few seconds later it _sometimes_ does eventually see the "transmit on" toggle late (but not the subsequent "transmit off") and then locks in transmit carrier mode (but no modulation).

[...]

PR #171, opened 2025-01-12
Fix for Serial Command Parser in RX mode

dkaukov:
- Introduced a state machine (matchedDelimiterTokensRx) to reliably detect command delimiters.
- Eliminated issues caused by partial or incorrect delimiters.
- Made delimiter detection non-blocking.
- Optimized processing to handle incoming serial data as quickly as possible. 

Fixes https://github.com/VanceVagell/kv4p-ht/issues/166.

### DIFF
--- a/microcontroller-src/kv4p_ht_esp32_wroom_32/kv4p_ht_esp32_wroom_32.ino
+++ b/microcontroller-src/kv4p_ht_esp32_wroom_32/kv4p_ht_esp32_wroom_32.ino
@@ -47,6 +47,7 @@ long lastSMeterReport = -1;
 #define DELIMITER_LENGTH 8
 const uint8_t COMMAND_DELIMITER[DELIMITER_LENGTH] = {0xFF, 0x00, 0xFF, 0x00, 0xFF, 0x00, 0xFF, 0x00};
 int matchedDelimiterTokens = 0;
+int matchedDelimiterTokensRx = 0;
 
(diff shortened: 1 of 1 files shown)


VanceVagell:
Thank you!

VanceVagell:
Hey Dmitry. I've merged this into the main branch and it has helped with the tx problems.

However, this caused a new bug. When squelch or filter settings are changed, and the user hits Done, the new values are not applied until the user resets the app. This means that the Android app sent the settings values to the ESP32 firmware, but the settings were not actually applied.

Since squelch is sent with TUNE_TO, moving to another frequency does apply the squelch (but it is not applied until they do this or restart the app).

Could you please take a look?

dkaukov:
Hey @VanceVagell ,  

Thanks for letting me know about the issue. I think this might be addressed in the changes from [[PR #168](https://github.com/VanceVagell/kv4p-ht/pull/168)](https://github.com/VanceVagell/kv4p-ht/pull/168), specifically in [[this commit](https://github.com/VanceVagell/kv4p-ht/pull/168/commits/32c903fc6cd930a6959f326a17d5b1d79442f382)](https://github.com/VanceVagell/kv4p-ht/pull/168/commits/32c903fc6cd930a6959f326a17d5b1d79442f382). Could you verify if the issue persists with these updates applied?

VanceVagell:
That worked, thank you!

PR #179, opened 2025-01-19
Android: Switch AudioTrack to the AudioFormat.ENCODING_PCM_16BIT

dkaukov:
This PR makes two key changes to improve audio playback and handling:  
1. Switches `AudioTrack` to use `AudioFormat.ENCODING_PCM_16BIT` for better device compatibility.  
2. Updates the wire transfer format to `int8_t` to address issues with Android USB drivers.

#### **Rationale**  
1. **Better Compatibility:**  
   Some devices don’t support PCM 8-bit , as documented [[here](https://developer.android.com/reference/android/media/AudioFormat#ENCODING_PCM_8BIT)](https://developer.android.com/reference/android/media/AudioFormat#ENCODING_PCM_8BIT). Switching to PCM 16-bit ensures consistent playback across devices.  
2. **USB Driver Fix:**  
   Android USB drivers can add zeros to audio chunks ([[issue](https://github.com/mik3y/usb-serial-for-android/issues/599)](https://github.com/mik3y/usb-serial-for-android/issues/599)). Using `int8_t`, where `0` represents silence, minimizes audio artifacts caused by this bug.


VanceVagell:
Please merge the latest changes in main into this and resolve any conflicts. The Android app DB has changed schema, so I can't easily install this to test it (my devices are using the new schema).

dkaukov:
@VanceVagell merged

VanceVagell:
I'm hearing audio artifacts as of this PR, please see attached test video. Listen at 0:16, 0:17, and 0:24. There is some popping.

It was subtle in that recording, but earlier I was hearing a lot of popping when first testing this. Reminds me of small buffer underruns, but not 100% sure that's what this is.

(Note that I merged with the latest in main again to test this, which adds UHF support. Please merge on your side as well.)

https://github.com/user-attachments/assets/953a9ef7-6c45-44f2-b3fb-3694b6e476a3

VanceVagell:
I retested on main without your PR and I still hear the clicking, so it's unrelated, was just a red herring.

Please merge once more and I think we're good to finalize this PR.

dkaukov:
@VanceVagell re-merged and tested. Seems ok to me. 
BTW This is was causing crash:
```
moduleTypeSetting.value.equals("UHF")
```

in java it is better to do:
```
   "UHF".equals(moduleTypeSetting.value)
```
to be null safe

Issue #180, opened 2025-01-20
[ESP32] Fine tune the ADC bias voltage


SmittyHalibut:
https://github.com/VanceVagell/kv4p-ht/blob/main/microcontroller-src/kv4p_ht_esp32_wroom_32/kv4p_ht_esp32_wroom_32.ino#L227

(Line number may change after I submit my PR here soon.)

138 was determined experimentally, but I'm not convinced its the right value to be in the center of the ADC's range.

Look at samples from the ADC, without any DC adjustments of the normal code, and see how close to the center we are.  Alternatively, look at the DC offset adjustment, see how far off it is.  Adjust this 138 value to get that DC offset as close to 0 as possible.

Actually, that's a good point.  We can do this dynamically.  Hmm.  Anyway, this is a nice-to-have.

dkaukov:
@SmittyHalibut 
```
#define DAC_PIN 26                   // GPIO26 is DAC_CHANNEL_2 (Bias voltage output)
#define ADC_PIN 34                   // GPIO34 is ADC1_CHANNEL_6 (Input-only ADC pin)
#define TARGET_ADC_VALUE 2048        // Midpoint for 12-bit ADC (ideal bias point)
#define TOLERANCE 5                  // Allowable error margin for ADC value
#define ADC_SAMPLES 100              // Number of ADC samples to average

#include <Arduino.h>

uint16_t average_adc() {
    uint32_t total = 0;
    for (int i = 0; i < ADC_SAMPLES; i++) {
        total += analogRead(ADC_PIN); // Read ADC
        delayMicroseconds(50);        // Short delay between reads
    }
    return total / ADC_SAMPLES;       // Return average ADC value
}

uint8_t calibrate_bias() {
    uint8_t low = 0, high = 255, mid;
    while (low <= high) {
        mid = (low + high) / 2;        // Binary search midpoint
        dacWrite(DAC_PIN, mid);        // Set DAC voltage
        delay(10);                     // Allow signal to stabilize
        int adc_value = average_adc(); // Get averaged ADC reading
        if (abs(adc_value - TARGET_ADC_VALUE) <= TOLERANCE) {
            return mid;                // Found suitable bias value
        } else if (adc_value < TARGET_ADC_VALUE) {
            low = mid + 1;             // Increase DAC value
        } else {
            high = mid - 1;            // Decrease DAC value
        }
    }
    return mid;                        // Return last tried value
}

void setup() {
    Serial.begin(115200);
[...]

Issue #175, opened 2025-01-16
[pcb] v2.0d updates


SmittyHalibut:
I'm ordering another batch of v2 boards and want to make a few minor updates:

* Move the Reset and Program buttons further apart.  My sausage fingers can't press the buttons at the same time.
* SA818 is too hard to solder. Make the pads larger.
* Can't solder the SA818 heatsink. Larger holes might help this.
* I removed a few DNP components that were for build options, but were never used. Just cleaned up the layout.
* Labeled the pins on the SA818 to make it easier to poke at it with a volt meter or oscope to test what's up.
* Add the ADC auto biasing circuit.

Also:
* I'm not going to populate the headers. They are far and away the tallest part on the board.

The Bad News(tm):
* I had to move the LEDs.  The NeoPixel is in the same place, but the Power and User LEDs

SmittyHalibut:
I'm also looking at biasing the ADC input from the radio.  See https://github.com/VanceVagell/kv4p-ht/issues/148 and related tickets.

SmittyHalibut:
ADC Bias stuff:
* Issue: https://github.com/VanceVagell/kv4p-ht/issues/148
  * Other issue that this should address: https://github.com/VanceVagell/kv4p-ht/issues/140
* Pull Request: https://github.com/VanceVagell/kv4p-ht/pull/159
* Original author, who did the debugging and designing of this issue: https://github.com/dkaukov

Summary:
## v1 hardware
v1 hardware has no ADC biasing.  Audio is DC coupled from radio to microcontroller. Not sure how this ever worked.  I can believe that the output from the radio is AC coupled, but the ADC input to the ESP32 is not likely to have any DC path to ground, power, or internal bias.  This leaves it floating and indeterminate.  This could be why some people have issues and others don't.

## v2 hardware
v2.0c hardware is biased in hardware to 0.57v (47k/10k voltage divider from +3.3v).  This was chosen to be (about) the middle of the ADC native range of about 0.1v to 1.1v.  The analog audio needs to be about 1v p-t-p.

Software to make this work:
* `ADC_ATTEN_DB_0` at https://github.com/VanceVagell/kv4p-ht/blob/main/microcontroller-src/kv4p_ht_esp32_wroom_32/kv4p_ht_esp32_wroom_32.ino#L174 instead of `ADC_ATTEN_DB_12`.
[...]

PR #181, opened 2025-01-20
Support for hardware versions. Added NeoPixel. #175

SmittyHalibut:
## ESP32:
Added firmware support for:
* Detecting hardware versions.
  * Note: Sense_V[PN] pins don't have internal pull-[up|down] resistors, so the intended method of detecting versions won't work.  Hi-Z reads the same as LOW.  Really, it means that we've gone from 9 detectable versions to only 4.  One of those versions will have to specify another method of detecting, probably using different GPIO pins.  Anyway. For now, we're ok.
* Added support for NeoPixel.  Radio's state (stopped, Tx, Rx with squelch open, or closed) are now indicated on the NeoPixel by color.
* Fixed: v2.0 moved the Squelch pin to a different GPIO. Now that we can detect the hardware version, this is now fixed and the code looks in the right place.

## PCB
* Easier to solder, larger pads on the 818.
* Added ADC auto bias from DAC2 (see #180 and related)
* Cleaned up some unused outlines.
* Spread apart Reset and Program buttons so my fat sausage fingers and press them both.

### DIFF
--- a/website-src/updates.html
+++ b/website-src/updates.html
@@ -44,6 +44,9 @@
 				<ul>
 					<li>Adds UHF support (requires UHF radio module and UHF lowpass filter)</li>
 					<li>Adds physical PTT button support (for PCB v2.0, which includes buttons)</li>
+					<li>Adds support for custom memory offsets (e.g. odd offsets)</li>
+					<li>Adds support for rx tones to open squelch in a memory</li>
+					<li>Adds option to skip a memory during scan</li>
 					<li>Numerous other bug fixes and small improvements</li>
 				</ul>
 				<h1>Jan 12, 2025: Android app update v1.6.7, ESP32 firmware v8</h1>


microcontroller-src/kv4p_ht_esp32_wroom_32/kv4p_ht_esp32_wroom_32.ino
```diff
@@ -24,6 +24,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #include <driver/i2s.h>
 #include <driver/dac.h>
 #include <esp_task_wdt.h>
+#include <Adafruit_NeoPixel.h>
```
VanceVagell, line 27: Could you please update the contribute.html file in the website-src to indicate that this library should be installed for people who want to compile the firmware in Arduino IDE? There's a bulleted list of other dependencies there, you can just add this.

VanceVagell:
Please merge with main, resolve conflicts (mainly around how firmware determines VHF/UHF), and test. When I did that, I get this error (seen in serial output from the ESP32):

Guru Meditation Error: Core  1 panic'ed (LoadProhibited). Exception was unhandled.

It then crashes and restarts the firmware, this happens on a loop.

SmittyHalibut:
> Please merge with main, resolve conflicts (mainly around how firmware determines VHF/UHF)


> ...and test. When I did that, I get this error (seen in serial output from the ESP32):
> 
> Guru Meditation Error: Core 1 panic'ed (LoadProhibited). Exception was unhandled.
> 
> It then crashes and restarts the firmware, this happens on a loop.

Are you seeing this when just connected to a terminal emulator?  Or is there a way to see what's coming over serial in the app?

microcontroller-src/kv4p_ht_esp32_wroom_32/kv4p_ht_esp32_wroom_32.ino
SmittyHalibut: Done.

SmittyHalibut:
> Please merge with main, resolve conflicts (mainly around how firmware determines VHF/UHF), and test. When I did that, I get this error (seen in serial output from the ESP32):
> 
> Guru Meditation Error: Core 1 panic'ed (LoadProhibited). Exception was unhandled.
> 
> It then crashes and restarts the firmware, this happens on a loop.

What do you see on the phone app when this happens?  I've got it saying "No firmware installed."  Is that what you're seeing?  I'm just trying to confirm whether I'm chasing a phantom or not.

VanceVagell:
Yes, it says no firmware installed (because the ESP32 isn't replying to the GET_VERSION command). To further investigate I reviewed the Arduino IDE terminal with the ESP32 connected and see the error I included above. I've never seen that error before, so I suspect it's something in this PR.

microcontroller-src/kv4p_ht_esp32_wroom_32/kv4p_ht_esp32_wroom_32.ino
```diff
@@ -730,3 +832,52 @@ void processTxAudio(uint8_t tempBuffer[], int bytesRead) {
     bytesToWrite -= bytesWritten;
   } while (bytesToWrite > 0);
 }
+
+uint8_t get_hardware_version() {
```
dkaukov, line 836: This seems too complicated to what it does. Is it effectively the same to the: 
```
uint8_t get_hardware_version() {
  pinMode(HW_VER_0_PIN, INPUT_PULLDOWN); pinMode(HW_VER_1_PIN, INPUT_PULLDOWN);
  uint8_t down = (digitalRead(HW_VER_1_PIN) << 1) | digitalRead(HW_VER_0_PIN);
  pinMode(HW_VER_0_PIN, INPUT_PULLUP); pinMode(HW_VER_1_PIN, INPUT_PULLUP);
  uint8_t up = (digitalRead(HW_VER_1_PIN) << 1) | digitalRead(HW_VER_0_PIN);
  return (down == up) ? down : 0xFF; // Return version or error
}
```
And if there no pull-ups support on those pins, we can do:
```
uint8_t get_hardware_version() {
  pinMode(HW_VER_0_PIN, INPUT_PULLDOWN); pinMode(HW_VER_1_PIN, INPUT_PULLDOWN);
  return (digitalRead(HW_VER_1_PIN) << 1) | digitalRead(HW_VER_0_PIN);
}
```

SmittyHalibut:
I've confirmed I'm fighting the fact that I don't have the most current app version on my phone.  That means I have to setup the Google app development environment, which I don't have the time to do today.  I've already spent longer on this PR than I intended today.

On the plus side, all my new code: NeoPixels, Hardware Versions, and the new SoftwareSerial console I've put on the GPIO header, are all `#ifdef` wrapped now so it should be a bit easier to narrow down where the problem is.  I've also got another serial console I can write to for `printf()` debugging (my favorite!)

Yeah, I just confirmed that comms are stuck at getting a firmware version.  I can't debug anymore without the current app version.  This will have to be later.

microcontroller-src/kv4p_ht_esp32_wroom_32/kv4p_ht_esp32_wroom_32.ino
```diff
@@ -116,6 +119,36 @@ int fadeDirection = 0; // 0: no fade, 1: fade in, -1: fade out
...
+//#define HW_V1 (0x22)  // NO! Sense pins can't pull-[up|down]. Always detect as LOW.
+
+#define HW_VER_1_PIN 36
+#define HW_VER_0_PIN 39
+uint8_t hardware_version = 0xFF;
+
+// Setup NeoPixel for debugging.
+#define PIXELS_NUM 1
+#define PIXELS_PIN 13
+Adafruit_NeoPixel pixels(PIXELS_NUM, PIXELS_PIN, NEO_GRB + NEO_KHZ400);
```
dkaukov, line 145: Is it possible to unit and call NeoPixels conditionally? For specific hardware revisions? Or turn on/off in settings?

microcontroller-src/kv4p_ht_esp32_wroom_32/kv4p_ht_esp32_wroom_32.ino
```diff
@@ -166,7 +245,12 @@ void initI2SRx() {
 
   // Initialize ADC
   adc1_config_width(ADC_WIDTH_BIT_12);
-  adc1_config_channel_atten(ADC1_CHANNEL_6, ADC_ATTEN_DB_12);
+  if (hardware_version == HW_V2_0C) {
```
dkaukov, line 248: Maybe it is better approach to change it in settings?

microcontroller-src/kv4p_ht_esp32_wroom_32/kv4p_ht_esp32_wroom_32.ino
```diff
@@ -299,7 +383,12 @@ void loop() {
             radioModuleStatus = RADIO_MODULE_FOUND;
           }
 
-          result = dra->volume(8);
+          if (hardware_version == HW_V2_0C) {
+            // v2.0c has a lower input ADC range.
+            result = dra->volume(4);
```
dkaukov, line 388: Maybe it is better approach to change it in settings?

dkaukov (approved):
Looks good to me with a few minor comments.

@VanceVagell, I believe features like VHF/UHF, NeoPixel support, volume, attenuation, etc., should be implemented in a consistent way. There are two possible approaches:

- The firmware automatically detects and sets them up.
- The user can configure them through the Settings.

Feel free to merge or request further changes as you see fit.

Note: I personally prefer @SmittyHalibut’s approach of auto-detection, especially since we have plenty of ESP32 pins available. However, since you’ve started implementing the VHF/UHF switch in the settings, I wanted to raise this as a concern. ;)

SmittyHalibut:
Withdrawing this PR so I can split it into multiple smaller, easier to test, PRs.  So far, see #199 and #201.

SmittyHalibut:
@dkaukov I'm not ignoring your comments.  As I rework the changes in #175 into smaller PRs, I'll be taking your comments into account.  Thank you.

microcontroller-src/kv4p_ht_esp32_wroom_32/kv4p_ht_esp32_wroom_32.ino
SmittyHalibut: I realized the "no pull-[up|down]" thing as I was filing the PR, after already making all the changes.  You're right that this can be WAY simpler now.  I'm just kicking myself for not looking at this before releasing v2.0c into the wild.  *sigh*

microcontroller-src/kv4p_ht_esp32_wroom_32/kv4p_ht_esp32_wroom_32.ino
```diff
@@ -524,18 +613,26 @@ void loop() {
       size_t samplesRead = bytesRead / 2;
 
       bool squelched = (digitalRead(SQ_PIN) == HIGH);
+      if (first_time) {
```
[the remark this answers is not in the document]
SmittyHalibut: Agreed.  After doing all the `#ifdef` stuff, I was very quickly regretting doing it all inline and not abstracting it out.  Now that I'm re-doing it all, I will do it better.

microcontroller-src/kv4p_ht_esp32_wroom_32/kv4p_ht_esp32_wroom_32.ino
SmittyHalibut: Yes, absolutely.  Doing it for specific hardware revisions isn't necessary since it doesn't hurt to squirt NeoPixel colors into a disconnected GPIO pin.  But allowing the user to disable the LEDs in software definitely makes sense.  I'll make that easier to do.

microcontroller-src/kv4p_ht_esp32_wroom_32/kv4p_ht_esp32_wroom_32.ino
SmittyHalibut: Why's that?  This is a hardware dependent change, dependent on where the bias voltage has been set.  v2.0c has a fixed bias at 0.56v, therefore requires a low output volume from the 818 so it doesn't clip into the negative, and therefore requires `DB_0`.  

v2.0d uses your variable bias design and therefore should use the maximum signal to noise ratio design.

SmittyHalibut:
> Note: I personally prefer @SmittyHalibut’s approach of auto-detection, especially since we have plenty of ESP32 pins available. However, since you’ve started implementing the VHF/UHF switch in the settings, I wanted to raise this as a concern. ;) Also, I’d personally prefer to encode 'features' instead of board versions, but that’s just my take on it.

The only reason that VHF/UHF is implemented as a user option is because the SA818s don't report which band they are.  There's no way to auto-detect this from the SA818 module.  We could implement a way to specify at assemble time: jumper here for VHF, there for UHF.  *shrug*  That's not a horrible idea.

dkaukov:
Hi @SmittyHalibut,

This is exactly what I mean! For example, we could short `GPIO12` to ground to indicate UHF, and include the VHF/UHF info in the VersionInfo reply so the Android app can switch automatically. Similarly, shorting `GPIO13 `to ground could indicate the presence of a NeoPixel. This way, we’re encoding hardware features directly instead of relying on board revisions. With all the spare ESP32 pins, it’s simple and effective!

SmittyHalibut:
With the expansion headers, there aren't as many "unused" pins as it seems.  We can definitely reclaim some of those from the headers if we want/need.  But there isn't a glut of unused pins.  That's why I originally put the first two version detect on the Sense pins; they were the only two that weren't already assigned.

I think we'll run out of pins VERY quickly if we tried to make a pin for every feature.  I believe we should keep with board versions.  #myOpinion

microcontroller-src/kv4p_ht_esp32_wroom_32/kv4p_ht_esp32_wroom_32.ino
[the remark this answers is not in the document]
SmittyHalibut: This is done in #207

PR #187, opened 2025-01-22
More robust command extractor state machine

dkaukov:
This PR improves the command extractor state machine by making it more memory efficient, preventing data loss in edge cases, and using a less common frame separator.

Changes:

- Optimized memory usage for better efficiency.
- Fixed issues where data could be lost in certain conditions.
- Extracted `RxStreamParser` to the separate class.
- Changed the frame separator to 0xDEADBEEFDEADBEEF to avoid conflicts with the audio stream.


VanceVagell:
Thanks for the PR, generally looks good I just need to test it a bit before merging. 

One question, could you please clarify how 0xDEADBEEFDEADBEEF is less likely to conflict with the audio stream than the current delimeter?

dkaukov:
The original 0x00ff00ff00ff00ff delimiter worked fine with unsigned 8-bit PCM. However, after switching to signed PCM, it became 0, -1, 0, -1, 0, -1, 0, -1, which has a higher chance of showing up in the audio stream. The new delimiter 0xDEADBEEFDEADBEEF is less likely to conflict because, well, it’s "Dead Beef"—a less common pattern in audio data ;)

PR #192, opened 2025-01-25
USB Serial driver patch

dkaukov:
This fixes lost packets from USB serial driver. 
The main discussion is here: https://github.com/mik3y/usb-serial-for-android/pull/614


VanceVagell:
Wow this looks really great, thanks Dmitry. I will test this locally tomorrow before merging.

PR #193, opened 2025-01-26
480mhz UHF frequency limit

dkaukov:
This. change increases upper limit of the UHF band to the 480Mhz.


microcontroller-src/platformio.ini
```diff
@@ -21,5 +21,5 @@ build_flags =
   -DARDUINO_EVENT_RUNNING_CORE=0
 lib_deps =
     ; fatpat/DRA818@^1.0.1 
-    https://github.com/fatpat/arduino-dra818.git#v1.0.1 ; v1.0.1 is the latest release, but has not been pushed to registry. Update later
+    https://github.com/fatpat/arduino-dra818.git
```
dkaukov, line 24: We need to upgrade to the latest.

VanceVagell:
No country allows amateur radio transmissions above 450MHz. The values in the drop-down currently let people select options that are compliant for where they live. If someone has an unusual reason to increase or change these, they are welcome to fork the project.

dkaukov:
Hi @VanceVagell 

Thanks for the feedback. I understand that transmissions above 450 MHz aren't allowed for amateur radio, and the dropdown is designed to ensure compliance. However, some users might have valid reasons to use frequencies up to 480 MHz, such as experimental licenses or non-amateur applications.

Also, why limit RX? Receiving signals above 470 MHz doesn’t break any rules and could be useful for monitoring or experimenting.

VanceVagell:
I see now that SA818-U can rx at 480MHz, so I've increased the rx max value in the app.

I had been looking at the DRA818-U datasheet, which stops at 470MHz.

dkaukov:
Hi @VanceVagell,
Thanks for updating the RX max value! That’s a good start, but it’s not the whole picture. We also need to tweak the firmware to handle the module properly:  

```cpp
if (paramsStr.charAt(0) == 'v') {
    dra = new DRA818(&Serial2, SA818_VHF);
} else if (paramsStr.charAt(0) == 'u') {
    dra = new DRA818(&Serial2, SA818_UHF);
}
```  

To make that work, we’ll need to upgrade to the latest version of the `arduino-dra818` library. That’ll bring everything in line with the new changes and make sure the hardware is fully supported.

PR #194, opened 2025-01-26
Startup crash fix

dkaukov:
With clean settings.


VanceVagell:
Thank you for catching and fixing this!

PR #197, opened 2025-01-27
Usb driver tweaks

dkaukov:
- fixed busywait (less battery drain)
- cancel and close pending USB requests on shutdown (better switch to flasher)


android-src/KV4PHT/app/src/main/java/com/vagell/kv4pht/radio/RadioAudioService.java
```diff
@@ -386,6 +387,19 @@ public void setMode(int mode) {
                 sendCommandToESP32(RadioAudioService.ESP32Command.STOP);
                 audioTrack.stop();
                 usbIoManager.stop();
+                try {
+                    serialPort.setDTR(false);
```
VanceVagell, line 391: Did you confirm that this works? In my testing, using setDTR and setRTS did not get the ESP32 to do what it should (I was trying to get it to go into bootloader mode).

dkaukov: It is just reset. Resetting works fine.

PR #201, opened 2025-01-28
Verify bytes read are actually delimiter.

SmittyHalibut:
Verify that the characters read are actually the delimiter characters, and reset the count if not.  This will re-sync communications between the ESP32 and the Android app if things get wonky.

Note that waiting on `Serial.available()` isn't necessary since `Serial.read()` already blocks until a character is available.  If you were looking for some other condition in that loop to exit it, it would be useful.  But since we aren't, I optimized it out.


microcontroller-src/kv4p_ht_esp32_wroom_32/kv4p_ht_esp32_wroom_32.ino
```diff
@@ -124,6 +124,10 @@ int fadeDirection = 0; // 0: no fade, 1: fade in, -1: fade out
 int attenuation = ATTENUATION_MAX; // Full volume
 bool lastSquelched = false;
 
+// 11dB vs 12dB is a ...version thing?
```
SmittyHalibut, line 127: Err... This is #199.  I suspect once it gets merged, this will update and go away.

VanceVagell:
Please merge with main.

Also, did you test this on actual hardware and confirm it works as expected?

SmittyHalibut:
> Please merge with main.
> 
> Also, did you test this on actual hardware and confirm it works as expected?

I tested on actual hardware and ... it didn't break?  That is, the sync at the top of `loop()` still syncs.

I _THINK_ this was the problem we were running into with #181, but I'm not able to verify.  I can't verify.  I can verify that the sync is still working, but I can't verify this fixes anything.  It is/should be safe belt-and-bracers sort of coding, IMHO.

PR #205, opened 2025-01-28
Add hardware detection. Fix always-open squelch on v2 boards.

SmittyHalibut:
There was some merge wackiness, but I think it's all good now.

Added detection of hardware version, and made some config changes based on hardware version:
* Squelch from SA818 moved to GPIO4. This is now properly detected.
* Adjust volumes for v2.0c, which has a different bias setting for receive audio.


microcontroller-src/kv4p_ht_esp32_wroom_32/kv4p_ht_esp32_wroom_32.ino
```diff
@@ -139,13 +151,34 @@ void tuneTo(float freqTx, float freqRx, int tone, int squelch, String bandwidth)
 void setMode(int newMode);
 void processTxAudio(uint8_t tempBuffer[], int bytesRead);
 void iir_lowpass_reset();
+hw_ver_t get_hardware_version();
+void reportPhysPttState();
```
SmittyHalibut, line 155: It looks like Arduino doesn't require forward declarations. C/C++ does, so PlatformIO/gcc does.  This is here so I could compile the previous commit from @VanceVagell

Issue #97, opened 2024-11-16
[Android] Add ESP32 stub loader support for much faster firmware flashing


VanceVagell:
FirmwareUtils and CommandInterfaceESP32 need to be updated to install the ESP32 stub loader into RAM, then increase the baud to 921600 and increase packet size once the stub loader is successfully installed.

The stub loader bytes are already bundled with the Android app in the raw/ directory, they just need to be installed.

VanceVagell:
Whomever works on this can start by seeing what I was trying to do, to see how it should roughly fit into the flow.

See FirmwareUtils.flashFirmware() where loadStubFromFile() is commented out, and see the unused implementation of that method. This does not work as-is, but might give a hint as to a working direction.

VanceVagell:
No longer needed as PR #206 addresses this!

PR #206, opened 2025-01-28
Auto boot loader

dkaukov:
- extended serial port to be able to set DTR and RTS together
- used this method to enter to the bootloader
- fixed drain(). The current serial driver actually rejecting small buffers
- cleaned up waiting for reply


android-src/KV4PHT/app/src/main/java/com/vagell/kv4pht/firmware/bearconsole/CommandInterfaceESP32.java
```diff
@@ -269,10 +271,10 @@ public void changeBaudRate() {
     }
 
     private void drain() {
-        byte[] buf = new byte[1];
+        byte[] buf = new byte[1024];
         while (true) {
             try {
-                int bytesRead = mSerialPort.read(buf, 1000);
+                int bytesRead = mSerialPort.read(buf, 5);
```
SmittyHalibut, line 277: Yikes.  Good catch.

SmittyHalibut (approved):
All the changes look good/reasonable to me.  No complaints from a style or visual semantic stand point.

I assume you've tested this and it correctly triggers a program of the ESP32?  I'm personally very interested in getting this to work correctly.  The hardware is there in v2, I want to use it, so the user doesn't have to take the board out of the case every time a new app comes out.

VanceVagell:
Can confirm this not only works to enter the bootloader automatically (at least on 1.x PCBs), but also fixes the flashing speed (resolved bug #97).

You deserve a gold medal for this one! Thank you!

Issue #203, opened 2025-01-28
[ESP32] Add Hardware Version detection


SmittyHalibut:
Read Sense_VP and Sense_VN, they're either HIGH or LOW.

| Sense_VP / GPIO36 | Sense/VN / GPIO39 | Hardware version |
|-------------------|-------------------|------------------|
| LOW | LOW | v1.x, v2.0b |
| LOW | HIGH | unused |
| HIGH | LOW | v2.0d |
| HIGH | HIGH | v2.0c |

The original intent was to also use Hi-Z as a detectable state by setting internal pull-ups and pull-downs, but the Sensor pins don't have output drive hardware.  So we're stuck with external drives.

The good news is, they are analog pins, so we can provide analog voltages with voltage dividers to provide more states going forward.  For now, we're just going to use HIGH and LOW.

SmittyHalibut:
Oh!  This also fixes the fact that squelch never closes on v2.0 hardware.  The squelch detect pin changed in v2.0; we need to detect the hardware version to make this software change.

SmittyHalibut:
Oops. This is a dupe of #122.

SmittyHalibut:
https://github.com/VanceVagell/kv4p-ht/pull/205 is the PR for this.

dkaukov:
Hi @SmittyHalibut @VanceVagell 
On the `1.7b` board, I had to add pull-down resistors on `GPIO36` and `GPIO39` to ground. Without them, the hardware version detection was unreliable—sometimes reporting incorrect values. It seems like floating inputs are causing intermittent issues, though the occurrence is quite rare. It might be worth updating the documentation and release notes to mention this for others who might run into the same problem.

SmittyHalibut:
#205 is done. Resolving.

PR #208, opened 2025-01-29
Add NeoPixel support.

SmittyHalibut:
This PR is built-upon #203.  So things will probably get wonky when #203 is merged, requiring a rebase of this one.


microcontroller-src/kv4p_ht_esp32_wroom_32/kv4p_ht_esp32_wroom_32.ino
```diff
@@ -556,17 +621,22 @@ void loop() {
...
 
-
       size_t bytesRead = 0;
       static uint16_t buffer16[I2S_READ_LEN];
       static uint8_t buffer8[I2S_READ_LEN];
       ESP_ERROR_CHECK(i2s_read(I2S_NUM_0, &buffer16, sizeof(buffer16), &bytesRead, 100));
       size_t samplesRead = bytesRead / 2;
 
       bool squelched = (digitalRead(SQ_PIN) == HIGH);
+      if (first_time) {
```
dkaukov, line 631: Can we avoid such flags like "first_time"? Why not just call it once from `setup()` ?

microcontroller-src/kv4p_ht_esp32_wroom_32/kv4p_ht_esp32_wroom_32.ino
```diff
@@ -129,23 +133,76 @@ bool lastSquelched = false;
...
+    case HW_VER_V2_0D:
+      color = pixels.Color(0, 0, 32);
+      SQ_PIN = 4;
+      break;
+    default:
+      // Unknown version detected. Indicate this some way?
+      break;
+  }
+
+  for (uint8_t ndx = 3; ndx; ndx--) {
```
dkaukov, line 197: Can we move it to `show_LEDs()`?

microcontroller-src/platformio.ini
```diff
@@ -22,4 +22,5 @@ build_flags =
 lib_deps =
     ; fatpat/DRA818@^1.0.1 
     https://github.com/fatpat/arduino-dra818.git#v1.0.1 ; v1.0.1 is the latest release, but has not been pushed to registry. Update later
-    plerup/EspSoftwareSerial@^8.2.0
\ No newline at end of file
+    plerup/EspSoftwareSerial@^8.2.0
+    Adafruit/Adafruit NeoPixel
```
dkaukov, line 26: Why not this one https://fastled.io/docs/ ?  Is Adafruit still using bit-banging?

dkaukov: Also looks like arduino has `neopixelWrite()` function built-in nowadays.

microcontroller-src/kv4p_ht_esp32_wroom_32/kv4p_ht_esp32_wroom_32.ino
SmittyHalibut: I'm happy to move it to a separate function, but `show_LEDs()` isn't the right pace for it.  This is showing what version of hardware has been detected.  `show_LEDs()` is reporting what Mode the software is in.

Do you really think that's necessary?

microcontroller-src/platformio.ini
SmittyHalibut: Honestly: it's what I knew from years past.  Is there a reason NOT to use it?  It's not like we need the performance of FastLED, we've got a single LED.

I like the idea of built-in functions, but from what little documentation I could find on `neopixelWrite()`, it relies on an underlying system like FastLED or Adafruit to do the actual writing.  If you have other documentation, please point me to it.

Unless there's a compelling reason otherwise, I'd like to keep this for now.  We can refactor it later if a new reason to do so comes up.

dkaukov: I get that you’re using Adafruit NeoPixel out of familiarity, but I see two potential issues:  

1. **Do we really need an extra library?** Since we’re only driving a single LED, `neopixelWrite()` seems like a simpler, dependency-free alternative. The fewer external libraries we rely on, the better.  
2. **Blocking vs. RMT-based approach:** `neopixelWrite()` uses the RMT peripheral on ESP32, which means it doesn’t block the main loop like bit-banged implementations. This is important to keep our timing-sensitive code running smoothly.  

Unless there’s a specific reason to stick with Adafruit NeoPixel, I think `neopixelWrite()` would be the better choice here. What do you think?

microcontroller-src/kv4p_ht_esp32_wroom_32/kv4p_ht_esp32_wroom_32.ino
dkaukov: @SmittyHalibut  Instead of using LEDs to indicate the hardware revision, why not include it in the VersionInfo response and display it in the UI? That way, we keep the LEDs focused on operational status, which seems like a better use of their limited signaling capability.

VanceVagell: Thanks for the PR! An earlier prototype of the project used a Neopixel (when I was getting the Adafruit QT Py microcontoller to work with it), and it's definitely a great quality of life improvement to be able to communicate more with it.

I agree with @dkaukov that we don't want to introduce delays in setup() or loop() unless absolutely necessary. In this case, communicate the version doesn't hit that bar. If you want to keep this in here for debug purposes (i.e. you expect to use this regularly while testing / update code) then please comment it out, like some of the other debug statements and options throughout the code.

I've already tested this as-is on a PCB 1.x, so I'm ready to merge once you fix this last thing.

Thanks!

SmittyHalibut:
Yall make compelling arguments. :-)* take the hardware revision display out of setup(). It’ll still detect there, but that will get used in show_LEDs(). All delay()s will be removed from all code. Instead, I’ll use millis() to calculate state over time for any blinking I want to do. So it’s completely non-blocking. * For the purposes of quick acceptance and merge, I may put off the replacement of the Adafruit library. If it’s a simple API swap, I’ll do it. But from what I read (but haven’t verified yet), neopixelWrite() uses an underlying library, like FastLED or Adafruit, for the actual IO. Plus: the function signature has you provide a pin, and RGB value, but not an LED number. So the API isn’t a straight 1:1. More in a bit. On Jan 30, 2025, at 4:51 AM, VanceVagell ***@***.***> wrote:﻿
@VanceVagell commented on this pull request.



In microcontroller-src/kv4p_ht_esp32_wroom_32/kv4p_ht_esp32_wroom_32.ino:
> +      SQ_PIN = 32;
+      break;
+    case HW_VER_V2_0C:
+      color = pixels.Color(32, 0, 0);
+      SQ_PIN = 4;
+      break;
+    case HW_VER_V2_0D:
+      color = pixels.Color(0, 0, 32);
+      SQ_PIN = 4;
+      break;
+    default:
+      // Unknown version detected. Indicate this some way?
+      break;
+  }
+
+  for (uint8_t ndx = 3; ndx; ndx--) {

Thanks for the PR! An earlier prototype of the project used a Neopixel (when I was getting the Adafruit QT Py microcontoller to work with it), and it's definitely a great quality of life improvement to be able to communicate more with it.
I agree with @dkaukov that we don't want to introduce delays in setup() or loop() unless absolutely necessary. In this case, communicate the version doesn't hit that bar. If you want to keep this in here for debug purposes (i.e. you expect to use this regularly while testing / update code) then please comment it out, like some of the other debug statements and options throughout the code.
I've already tested this as-is on a PCB 1.x, so I'm ready to merge once you fix this last thing.
Thanks!

SmittyHalibut:
Ok.  So.  I spent all afternoon trying to find a bug in `neopixelWrite()` that I still haven't found, but it appears to only affect start-up, showing the hardware version, which is a nice-to-have anyway.  So I'm submitting my code anyway.

Changes:
* Switched from Adafruit NeoPixel to `neopixelWrite()` even though it won't show Blue until after `loop()` is running and I can't figure out why.
* Took all `delay()`s out of my code.  Instead, we are either showing a solid color, or "breathing" asynchronously, calculating the current state based on `millis()`.

SmittyHalibut:
Known issue:  In `setup()` we detect the hardware version and set the color we display in a STOPPED state.  v1 is green, v2.0c is red, v2.0d is blue.  (These happen to coincide with the color of the PCBs that I've fabricated, but that's arbitrary.)

Green displays correctly.  Red displays correctly.  Blue does not display correctly.

HOWEVER.  In all cases, when the code starts up enough to go into MODE_RX, blue DOES display correctly.

I've debugged everything as best I can; in all cases, the blue color value is being correctly passed to `neopixelWrite()`.  All cases use the same code path, though `show_LEDs() -> neopixelColor() -> neopixelWrite()`.

But in MODE_RX, the "Squelch is closed" color is blue, and that shows correctly.

I haven't figured out what happens between `case MODE_STOPPED` and `case MODE_RX` that kicks the library so it correctly sends a blue value to the LED.  I'm GUESSING it's a timer somewhere, but that's just speculation.

I also haven't hooked up a scope to verify whether the blue values are actually being sent or not.

But I decided this wasn't a release blocking bug, so here's the PR submitted for your approval, with everything else working as expected.  (At least, in my testing.)

SmittyHalibut:
https://github.com/VanceVagell/kv4p-ht/issues/210

microcontroller-src/kv4p_ht_esp32_wroom_32/kv4p_ht_esp32_wroom_32.ino
```diff
@@ -829,3 +852,61 @@ hw_ver_t get_hardware_version() {
...
+  static uint32_t next_time = 0;
+  const static uint32_t update_every = 50;    // in milliseconds
+
+  uint32_t now = millis();
+
+  // Only change LEDs if:
+  // * it's been more than update_every ms since we've last set the LEDs.
+  // * either mode or lastSquelched have changed.
+  // This allows us to call show_LEDs() as often as we want without risking slamming the micro with IO.
+  if (now >= next_time || last_mode != mode || last_squelched != lastSquelched) {
```
dkaukov, line 887: @SmittyHalibut looks good. 
Is tracking state changes necessary here? A 50ms update rate is already below human perception (~50-100ms), so we could simplify by updating LEDs every 50ms without extra checks.

microcontroller-src/kv4p_ht_esp32_wroom_32/kv4p_ht_esp32_wroom_32.ino
```diff
@@ -767,28 +792,26 @@ void tuneTo(float freqTx, float freqRx, int txTone, int rxTone, int squelch, Str
...
     case MODE_TX:
       txStartTime = micros();
-      digitalWrite(LED_PIN, HIGH);
       digitalWrite(PTT_PIN, LOW);
       initI2STx();
       txCachedAudioBytes = 0;
       isTxCacheSatisfied = false;
       break;
   }
+  show_LEDs();
```
dkaukov, line 814: I think, , show_LEDs() should be called once per loop to keep execution predictable.

microcontroller-src/kv4p_ht_esp32_wroom_32/kv4p_ht_esp32_wroom_32.ino
SmittyHalibut: It's a good question.  My PREFERENCE would be to keep both conditions, but it's not a hill I'd die on.

They are for different purposes.  The time based one is for animation, the state change is for ... state changes.  At 50ms for animation, it can serve both purposes.  But if the animation was 500ms, it would add some inappropriate latency.

Are we likely to ever increase from 50 to 500ms?  Not really.  But logically, they are different.  I don't think the cost of all three conditionals is particularly high.

dkaukov (approved):
LGTM

dkaukov:
Hi @SmittyHalibut, I took the liberty of cleaning this up for you: https://github.com/VanceVagell/kv4p-ht/pull/208/commits/0e0054e3ffb36bc37424001d96b8b1b7f60289b7. This also appears to fix the issue you mentioned.

VanceVagell:
@dkaukov what's the best way to handle that clean-up commit? Should @SmittyHalibut merge it into this PR, and then I merge this into main? Thanks for the help improving this.

dkaukov:
@VanceVagell @SmittyHalibut merged.

SmittyHalibut:
Consider giving the max value for `bright` which is 255:
```
(255 * 255) >> 8 = 254
(1 * 255) >> 8 = 0
```
Losing one brightness level from 255 to 254 is *shrug* no problem.  Going from 1 to 0 makes a huge difference.  I also like the efficiency of right shifting for division, but we can afford an integer division here.  It's still way better than floating point.  :-)

android-src/KV4PHT/app/src/main/res/raw/v11_kv4p_ht_esp32_wroom_32_ino.bin
VanceVagell, line 1: Please remove this file from the PR. I'll add new firmware binaries for each release, it's ok if the built-in firmware is incorrect while we're developing the next release as developers should be flashing the latest version directly.

VanceVagell:
I tested on 1.x hardware and there were no problems.

Once the one comment I added is resolved I can merge this. Thanks for the PR!

android-src/KV4PHT/app/src/main/res/raw/v11_kv4p_ht_esp32_wroom_32_ino.bin
SmittyHalibut: Done. Please merge when you're ready; I'm on the road for the next several days.

PR #209, opened 2025-01-31
Proposal for delivering debug messages from firmware

dkaukov:
This PR introduces a way to send debug messages from firmware for better observability with minimal overhead.

Key Questions

- Is this useful? Does structured debug messaging provide enough value?
- Debug vs. Release Builds? Should we limit debug messages to debug builds?
- Logging Destination? Is logcat the right place, or should we use something else?

### DIFF
--- a/android-src/KV4PHT/app/src/main/java/com/vagell/kv4pht/radio/RadioAudioService.java
+++ b/android-src/KV4PHT/app/src/main/java/com/vagell/kv4pht/radio/RadioAudioService.java
@@ -151,9 +151,14 @@ public class RadioAudioService extends Service {
 
     // Delimiter must match ESP32 code
     static final byte[] COMMAND_DELIMITER = new byte[] {(byte)0xDE, (byte)0xAD, (byte)0xBE, (byte)0xEF, (byte)0xDE, (byte)0xAD, (byte)0xBE, (byte)0xEF};
-    private static final byte COMMAND_SMETER_REPORT = 0x53; // Ascii "S"
-    private static final byte COMMAND_PHYS_PTT_DOWN = 0x44; // Ascii "D"
-    private static final byte COMMAND_PHYS_PTT_UP = 0x55;   // Ascii "U"
+    private static final byte COMMAND_SMETER_REPORT  = 0x53; // Ascii "S"
+    private static final byte COMMAND_PHYS_PTT_DOWN  = 0x44; // Ascii "D"
+    private static final byte COMMAND_PHYS_PTT_UP    = 0x55;   // Ascii "U"
+    private static final byte COMMAND_DEBUG_INFO     = 0x01;
+    private static final byte COMMAND_DEBUG_ERROR    = 0x02;
+    private static final byte COMMAND_DEBUG_WARN     = 0x03;
+    private static final byte COMMAND_DEBUG_DEBUG    = 0x04;
+    private static final byte COMMAND_DEBUG_TRACE    = 0x05;
 
     private final ESP32DataStreamParser esp32DataStreamParser = new ESP32DataStreamParser(this::handleParsedCommand);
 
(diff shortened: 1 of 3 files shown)


dkaukov:
image

dkaukov:
image

android-src/KV4PHT/app/src/main/java/com/vagell/kv4pht/radio/RadioAudioService.java
```diff
@@ -945,8 +950,8 @@ public void onRunError(Exception e) {
             }
         });
         usbIoManager.setWriteBufferSize(90000); // Must be large enough that ESP32 can take its time accepting our bytes without overrun.
-        usbIoManager.setReadBufferSize(1024/4); // Must not be 0 (infinite) or it may block on read() until a write() occurs.
-        usbIoManager.setReadBufferCount(16*4);
+        usbIoManager.setReadBufferSize(1024); // Must not be 0 (infinite) or it may block on read() until a write() occurs.
```
VanceVagell, line 953: This change seems unrelated. Is it intentional?

microcontroller-src/kv4p_ht_esp32_wroom_32/kv4p_ht_esp32_wroom_32.ino
```diff
@@ -112,8 +117,8 @@ char radioModuleStatus            = RADIO_MODULE_NOT_FOUND;
 bool i2sStarted = false;
 
 // I2S audio sampling stuff
-#define I2S_READ_LEN 1024
-#define I2S_WRITE_LEN 1024
+#define I2S_READ_LEN 1024/4
```
VanceVagell, line 120: Similar question. Seems unrelated, is this intentional? If so please clarify what these changes are doing.

microcontroller-src/kv4p_ht_esp32_wroom_32/kv4p_ht_esp32_wroom_32.ino
```diff
@@ -591,17 +647,19 @@ void loop() {
...
         lastSMeterReport = millis();
+      
       }
 
 
       size_t bytesRead = 0;
       static uint16_t buffer16[I2S_READ_LEN];
       static uint8_t buffer8[I2S_READ_LEN];
-      ESP_ERROR_CHECK(i2s_read(I2S_NUM_0, &buffer16, sizeof(buffer16), &bytesRead, 100));
+      ESP_ERROR_CHECK(i2s_read(I2S_NUM_0, &buffer16, sizeof(buffer16), &bytesRead, 0));
```
VanceVagell, line 662: What's the purpose of this change?

VanceVagell (changes requested):
Please see open comments with questions. Overall looks good and this will be really helpful for debugging certain issues. Thanks!

android-src/KV4PHT/app/src/main/java/com/vagell/kv4pht/radio/RadioAudioService.java
dkaukov: This was a trade-off in the approach. To handle traffic spikes caused by debug messages, I had to increase the receive buffers. However, I was stress testing it by sending a message in every loop, so it might not be necessary with a more "careful" use. Open to adjusting if needed.

microcontroller-src/kv4p_ht_esp32_wroom_32/kv4p_ht_esp32_wroom_32.ino
dkaukov: Since we started measuring the timing of the main loop, I couldn’t resist optimizing it. Reducing the read size and removing the timeout prevents the main loop from blocking on I2S read and UART write. As a result, the loop frequency increased from 20Hz to 5kHz.

microcontroller-src/kv4p_ht_esp32_wroom_32/kv4p_ht_esp32_wroom_32.ino
dkaukov: see next comment.

dkaukov:
@VanceVagell, this might be a good opportunity to start splitting the codebase into separate modules. I can extract the debug macros and metric collector into a separate header file to improve organization and maintainability. Let me know what you think!

VanceVagell:
If you'd like to submit a separate PR that cleans up and refactors the code for better organization in separate files, that would be great. But not super high priority. I'd like to do an overall refactoring of the firmware code to remove code duplication, standardize the kinds of command codes (some are characters, some are actual bytes, etc.), and have a cleaner state machine. That's all in this same universe of making the firmware more maintainable. Thanks for the help!

PR #216, opened 2025-01-31
Fix SA818 Module Definition and Improve Squelch Behavior

dkaukov:
### Summary  
This PR corrects the module definition and removes redundant code related to squelch (SQ) fade-in/fade-out, which was causing an unintended **loud pop** on startup.  

### Changes  
1. **Module Correction:**  
   - Updated the module reference from `DRA818` to `SA818` to accurately reflect the hardware in use.  

2. **SQ Fade-in/Fade-out Fix:**  
   - Removed redundant logic related to SQ fade-in/fade-out, which was not needed and introduced unintended behavior.  
   - This eliminates a **loud pop/bang** occurring immediately after a cold start.

### DIFF
--- a/.github/workflows/arduino.yml
+++ b/.github/workflows/arduino.yml
@@ -30,5 +30,6 @@ jobs:
           libraries: |
             - name: EspSoftwareSerial
               version: 8.1.0
-            - source-url: https://github.com/fatpat/arduino-dra818.git#89582e3ef7bf3f31f1af149e32cec16c4b9e4cf2
+            - source-url: https://github.com/fatpat/arduino-dra818.git
+              vaersion: 89582e3ef7bf3f31f1af149e32cec16c4b9e4cf2
           verbose: true
--- a/.github/workflows/arduino.yml
+++ b/.github/workflows/arduino.yml
@@ -30,5 +30,5 @@ jobs:
           libraries: |
             - name: EspSoftwareSerial
               version: 8.1.0
-            - source-url: https://github.com/fatpat/arduino-dra818.git@89582e3ef7bf3f31f1af149e32cec16c4b9e4cf2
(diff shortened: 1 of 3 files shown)


.github/workflows/arduino.yml
```diff
@@ -30,5 +30,5 @@ jobs:
           libraries: |
             - name: EspSoftwareSerial
               version: 8.1.0
-            - source-url: https://github.com/fatpat/arduino-dra818/archive/refs/tags/v1.0.1.zip
+            - source-url: https://github.com/fatpat/arduino-dra818.git
```
wsciaroni, line 33: This seems reasonable to me. We specify the commit/tag here. This will ensure we always work (and won't get broken by a new release of dra818 until we manually update to it).

wsciaroni (changes requested):
This looks like a good improvement. I have not tested it, but the code looks good to me.

Thank you for your contribution!

.github/workflows/arduino.yml
dkaukov: Hi, @wsciaroni 

That makes sense. However, the [`[arduino-dra818](https://github.com/fatpat/arduino-dra818)`](https://github.com/fatpat/arduino-dra818) library doesn’t have any release tags, and the last commit was over two years ago. Given its inactivity, we’re fairly safe using the master branch as a reference for now.  

If we want a more robust approach, we could fork the repository and reference a specific commit from our fork. Alternatively, reaching out to the maintainers to request a tagged release might be an option, but given the lack of activity, we may not get a response.

wsciaroni: Even though it is inactive, let's point to the commit hash of the latest commit on the default branch. That way it's very clear what version we're using when "future we" are debugging a weird problem.

dkaukov: @wsciaroni Got it! I don’t think it’s really needed, but I’ve updated it to the latest commit hash anyway.

Issue #227, opened 2025-02-07
[Android, ESP32] Increase time before Android app gives up on hearing back from ESP32 with version


VanceVagell:
Right now we give the ESP32 firmware a few seconds to reply to the Android app's version request command. If it times out, the app assumes there is no firmware. On slow devices, it's possible this could lead to a false negative where the ESP32 firmware just took a little too long.

Here's the change we should make:

1) As soon as app starts, show a persisent "Connecting to radio..." snackbar
2) If the radio is connected and firmware replies, close that snackbar without further notification

If there's any error trying to connect to the firmware, including timing out, hide the connecting snackbar and show the error.

Then, increase the connection timeout to a longer time such as 8 seconds or so (just below watchdog timer in ESP32 app).

Issue #226, opened 2025-02-07
[Android, ESP32] Radio module detection error doesn't seem to be working


VanceVagell:
A user came to Discord with a device that kept asking to be re-flashed, despite flashing successfully. It turns out their SA818 radio module was not yet installed. 

This should have resulted in the error "Radio module not responding to ESP32, check PCB solder joints" but instead was showing "kv4p HT radio not found, plugged in?"

This is the relevant code in the Android app:
https://github.com/VanceVagell/kv4p-ht/blob/8d4737c84fe633fbf447a8c0fc0b6fa91771a403/android-src/KV4PHT/app/src/main/java/com/vagell/kv4pht/radio/RadioAudioService.java#L1313

And the relevant code in the firmware:
https://github.com/VanceVagell/kv4p-ht/blob/8d4737c84fe633fbf447a8c0fc0b6fa91771a403/microcontroller-src/kv4p_ht_esp32_wroom_32/kv4p_ht_esp32_wroom_32.ino#L440

Not sure how/when this broke, Can easily test this by flashing an ESP32 with the firmware (outside a PCB), which the app should recognize but display the correct message since it can't contact the radio module.

PR #230, opened 2025-02-11
Improve Handshake Process with HELLO Message

dkaukov:
Currently, the host waits for an arbitrary delay after a USB connect event before sending a GET_VERSION request to determine if the module is ready. This can lead to timing issues and unreliable initialization.

This PR updates the handshake process so that the module proactively sends a HELLO message after boot. This ensures the host knows exactly when the module is ready, improving reliability and reducing unnecessary delays.

Changes:

1. The module now sends a HELLO message immediately after boot.
2. The host listens for the HELLO message before proceeding with further communication.
3. If no HELLO is received within a timeout period, the host prompts for flash

Benefits:

1. Eliminates reliance on arbitrary delays.
2. Ensures the host only communicates when the module is actually ready.
3. Improves reliability, especially in cases where boot time varies.
4. allows to re-sync after module reboot in RX mode

### DIFF
--- a/android-src/KV4PHT/app/src/main/java/com/vagell/kv4pht/radio/RadioAudioService.java
+++ b/android-src/KV4PHT/app/src/main/java/com/vagell/kv4pht/radio/RadioAudioService.java
@@ -1035,7 +1035,7 @@ public class RadioAudioService extends Service {
                 callbacks.missingFirmware();
                 setMode(MODE_BAD_FIRMWARE);
             }
-        }, 30000);
+        }, 60000);
     }
 
     private void initAfterESP32Connected() {
--- a/android-src/KV4PHT/app/src/main/java/com/vagell/kv4pht/radio/RadioAudioService.java
+++ b/android-src/KV4PHT/app/src/main/java/com/vagell/kv4pht/radio/RadioAudioService.java
@@ -979,7 +979,7 @@ public class RadioAudioService extends Service {
                 callbacks.missingFirmware();
                 setMode(MODE_BAD_FIRMWARE);
             }
-        }, 500);
+        }, 1000);
     }
 
     /**
--- a/android-src/KV4PHT/app/src/main/java/com/vagell/kv4pht/radio/RadioAudioService.java
+++ b/android-src/KV4PHT/app/src/main/java/com/vagell/kv4pht/radio/RadioAudioService.java
@@ -211,6 +211,7 @@ public class RadioAudioService extends Service {
     private boolean radioModuleNotFound = false;
     private boolean checkedFirmwareVersion = false;
     private boolean gotHello = false;
+    private final Handler timeOutHandler = new Handler(Looper.getMainLooper());
 
     // Safety constants
     private static int RUNAWAY_TX_TIMEOUT_SEC = 180; // Stop runaway tx after 3 minutes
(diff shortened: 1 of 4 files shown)


VanceVagell:
I'm confused, doesn't this just move the timer to waiting for HELLO instead of waiting for the GET_VERSION response?

Also, the microcontoller won't restart when the kv4p HT app is closed and reopened, meaning the HELLO message won't be sent, right?

Maybe I'm missing something...

dkaukov:
Hi @VanceVagell 

1. **Doesn’t this just move the timer to waiting for HELLO instead of GET_VERSION?**
Not really—right now, we’re guessing when the module is ready. With HELLO, the module tells us exactly when it's up, so we don’t waste time waiting too long or retrying too soon. No guessing, just a direct signal.

2. **What if the app restarts but the microcontroller doesn’t?**
Not a problem! The microcontroller will restart anyway (watchdog takes care of that). And to speed things up, we can force a restart right after USB connects. That way, HELLO always gets sent, keeping things reliable.

dkaukov:
May fix: https://github.com/VanceVagell/kv4p-ht/issues/227, https://github.com/VanceVagell/kv4p-ht/issues/226

VanceVagell:
Can you please confirm if this fixes bug #226 ? You can test it by plugging into your phone a bare ESP32 (not in a kv4p HT PCB or connected to a radio module) with the firmware installed.

If so I'll merge this in, otherwise it's not clear to me that this has added benefit over current functionality (I appreciate your explanation, and I can go into more detail if this does not fix #226).

dkaukov:
Hi @VanceVagell,

Yes, I can confirm that this does fix https://github.com/VanceVagell/kv4p-ht/issues/226. I tested it by plugging a bare ESP32 (not in a kv4p HT PCB or connected to a radio module) into my phone with the firmware installed, and it worked as expected.

android-src/KV4PHT/app/src/main/java/com/vagell/kv4pht/radio/RadioAudioService.java
```diff
@@ -958,19 +969,17 @@ public void onRunError(Exception e) {
...
-        handler.postDelayed(new Runnable() {
-            @Override
-            public void run() {
-                if (!checkedFirmwareVersion) {
-                    checkFirmwareVersion();
-                }
+        timeOutHandler.removeCallbacksAndMessages(null);
+        timeOutHandler.postDelayed(() -> {
+            if (!gotHello) {
+                Log.d("DEBUG", "Error: Did not HELLO from module.");
```
VanceVagell, line 978: Nit: "Dit not receive HELLO" (missing word "receive"). This can be in a follow up PR.

android-src/KV4PHT/app/src/main/java/com/vagell/kv4pht/ui/MainActivity.java
```diff
@@ -1726,6 +1737,21 @@ private void showUSBSnackbar() {
         usbSnackbar.show();
     }
 
+    private void showHandshakeSnackbar() {
+        CharSequence snackbarMsg = "kv4p HT radio handshake.";
```
VanceVagell, line 1741: Android app users are not likely to know what this means. Please change the message to:

"Connecting to radio..."

VanceVagell:
Awesome, thanks for fixing that bug! I will merge this once you adjust the 2 strings I commented on.

android-src/KV4PHT/app/src/main/java/com/vagell/kv4pht/ui/MainActivity.java
dkaukov: done

android-src/KV4PHT/app/src/main/java/com/vagell/kv4pht/radio/RadioAudioService.java
dkaukov: done

PR #240, opened 2025-02-28
Refactor ESP32 to Host Stream

dkaukov:
This PR introduces two key changes:

1. **Wraps Audio Stream into a Frame**: Audio data is now encapsulated into frames, streamlining the data structure.
2. **Adheres to Standard Protocol for GET_VERSION Reply**: The GET_VERSION response now follows the standard protocol format: `[DELIMITER(8 bytes)] [CMD(1 byte)] [paramLen(1 byte)] [param data(N bytes)]`.

With these changes, the protocol is unified in one direction, simplifying the parser by eliminating the need for the lookahead buffer. If this PR is accepted, I will proceed with converting the "Host to ESP32" stream to align with this protocol as well.


VanceVagell:
This is a good step in the right direction in cleaning up the firmware and messaging protocol, thanks Dmitry!

PR #241, opened 2025-03-02
Wrap Tx Audio Stream into Frames & Unify Host-to-ESP32 Protocol

dkaukov:
This PR builds on (https://github.com/VanceVagell/kv4p-ht/pull/240), introducing a structured protocol for transmitting audio and commands:  

```
[DELIMITER (8 bytes)] [CMD (1 byte)] [paramLen (1 byte)] [param data (N bytes)]
```

### **Key Changes:**  
- Unified protocol between Host and ESP32.  
  - Commands as enums.  
  - Parameters as packed structs (C++) and classes (Java).  
  - Sender + `FrameParser` on both sides.  
- Improved modularization with new modules.  
- Reduced Java boilerplate using Lombok.  

This makes communication more consistent, modular, and maintainable. 🚀


VanceVagell:
This is awesome, thanks! I'm assuming you tested this on hardware.

PR #244, opened 2025-03-03
De-bounce touch PTT button.

dkaukov:
By delaying touch-off event


dkaukov:
Hi @VanceVagell , 
It seems like it is working as expected for me, but you would need to double-check.

VanceVagell:
@dkaukov this particular PR is fine, I tested it. HOWEVER, there were major bugs in prior PRs that I discovered while testing this. In particular, neither the tx frequency nor the tx tone was being properly set during tuneToMemory. I will submit a fix for those issues momentarily.

Going forward, could you please post a short video of testing your PRs so I can review and merge them more quickly, with confidence that they don't have significant issues?

Thanks for these improvement, it's nice to see the snappier firmware and better-defined command structure.

PR #250, opened 2025-03-18
Updated protocol doc

dkaukov:


dkaukov:
Hi @wsciaroni,
Thank you for the feedback!  

Regarding unique command IDs across `esp->phone` and `phone->esp`, I see your point, but I believe maintaining uniqueness across two separate enums adds unnecessary complexity without significant benefit. That said, if you feel strongly about it, PRs are always welcome!  

I've adopted the rest of your suggestions:  
- Added byte size information for each struct field.  
- Documented byte endianness and bit significance.

wsciaroni (approved):

PR #251, opened 2025-03-26
Boards to test Mini circuits filters

SmittyHalibut:
Some PCBs to test using MiniCircuits filters instead of discrete components.  Let's see how they compare to the v1.8 and v2.0 filter designs.  More specifically, there's one for UHF.


SmittyHalibut:
Abused my privilege and merged this without anyone else's review.  :-)

These are only new files, nothing existing has changed.

PR #252, opened 2025-03-27
Use Arduino Audio Tools for ADC/DAC and Transmit Compressed OPUS Audio over the wire

dkaukov:
- Integrated [[Arduino Audio Tools](https://github.com/pschatzmann/arduino-audio-tools)] for handling ADC and DAC operations.  
- Implemented [OPUS](https://opus-codec.org/) compression for transmitting audio over the wire, improving bandwidth efficiency. 
- Added lightweight [flow control](https://medium.com/coderscorner/http-2-flow-control-77e54f7fd518) for `HostToEsp32` stream.
- Serial speed now 115200.
- Uses native Java OPUS codec implementation on Android side
- `AudioRecord` now `AudioFormat.ENCODING_PCM_FLOAT` to avoid format conversions.
- Modified protocol to support frames bigger than 255 bytes. Currently MTU is set to 2048 bytes.

This change requires thorough testing to verify audio quality, latency, and reliability across different conditions.


VanceVagell:
I see you requested review on this but it seems to still be under development. Is this ready for merge?

dkaukov:
Hi @VanceVagell ,
Yes, this is ready for review and merge. If we want to move in this direction, it's better to do so soon, as merging changes from master later will be difficult.

.github/workflows/arduino.yml
```diff
@@ -32,4 +32,8 @@ jobs:
               version: 8.1.0
             - source-url: https://github.com/fatpat/arduino-dra818.git
               version: 89582e3ef7bf3f31f1af149e32cec16c4b9e4cf2
+            - source-url: https://github.com/pschatzmann/arduino-audio-tools/archive/refs/tags/v1.0.1.zip
```
VanceVagell, line 35: Please update https://github.com/VanceVagell/kv4p-ht/blob/main/other-licenses.txt with a copy-paste of the license text from all new open source projects we introduce to kv4p HT.

android-src/KV4PHT/app/src/main/java/com/vagell/kv4pht/radio/Protocol.java
```diff
@@ -16,13 +19,13 @@ public final class Protocol {
     private static final String TAG = Protocol.class.getSimpleName();
 
     // Delimiter must match ESP32 code
-    static final byte[] COMMAND_DELIMITER = new byte[]{(byte) 0xDE, (byte) 0xAD, (byte) 0xBE, (byte) 0xEF, (byte) 0xDE,
-        (byte) 0xAD, (byte) 0xBE, (byte) 0xEF};
+    static final byte[] COMMAND_DELIMITER = new byte[]{(byte) 0xDE, (byte) 0xAD, (byte) 0xBE, (byte) 0xEF};
```
VanceVagell, line 22: Why was the delimiter shortened? Is there less risk of this occuring in random over-the-wire bytes now for some reason?

android-src/KV4PHT/app/src/main/java/com/vagell/kv4pht/radio/Protocol.java
```diff
@@ -301,6 +354,9 @@ private void processByte(byte b) {
                 matchedDelimiterTokens++;
             } else if (matchedDelimiterTokens == COMMAND_DELIMITER.length + 1) {
                 commandParamLen = b & 0xFF;
+                matchedDelimiterTokens++;
+            } else if (matchedDelimiterTokens == COMMAND_DELIMITER.length + 2) {
```
VanceVagell, line 358: What is this new if-statement doing vs. the previous if-statement above? Not obvious to me why one is +1 and one is +2, might need a comment here.

microcontroller-src/kv4p_ht_esp32_wroom_32/kv4p_ht_esp32_wroom_32.ino
```diff
@@ -206,28 +214,27 @@ void handleCommands(RcvCommand command, uint8_t *params, size_t param_len) {
...
       }
-      lastRssiReport = millis();
     }
+    END_EVERY_N_MILLISECONDS();
   }
 }
 
 void loop() {
-  squelched = (digitalRead(SQ_PIN) == HIGH);
+  squelched = squelchDebounce.debounce((digitalRead(SQ_PIN) == HIGH));
```
VanceVagell, line 237: I don't think this needs to be debounced, it's reading the pin directly from the radio module, not a physical button.

android-src/KV4PHT/app/src/main/java/com/vagell/kv4pht/radio/Protocol.java
VanceVagell, line 1: I believe the protocol decoumentation here also needs to be updated. It could be in a follow up PR (maybe just delete it in this PR if you aren't ready to update it yet, so it's not misleading): https://github.com/VanceVagell/kv4p-ht/tree/main/microcontroller-src/kv4p_ht_esp32_wroom_32#readme

microcontroller-src/kv4p_ht_esp32_wroom_32/buttons.h
```diff
@@ -20,33 +20,21 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
 boolean isPhysPttDown     = false;
+Debounce pttDebounce(DEBOUNCE_MS);
 
 void inline buttonsSetup() {
-  // Optional physical PTT buttons
   pinMode(PHYS_PTT_PIN1, INPUT_PULLUP);
   pinMode(PHYS_PTT_PIN2, INPUT_PULLUP);
 }
 
 void inline buttonsLoop() {
```
VanceVagell, line 34: I'm not sure exactly what code is causing it, but there's a bug with this and PCB v2. To repro:

1) Start the kv4p HT app fresh (kill it and restart it if needed)
2) Press the physical PTT button on the kv4p HT
3) Tx starts as expected
4) Release the physical PTT button

Expected:
5) Tx stops

Actual;
5) Tx continues

To stop tx, you  need to press the software PTT button. It seems that after this initial failure, subsequent PTT pressures with the physical button work (I didn't test this too exhaustively). Please test this out and see how we can fix this.

android-src/KV4PHT/app/src/main/java/com/vagell/kv4pht/radio/Protocol.java
dkaukov: OPUS is sensitive to frame size, and our frames are slightly larger than 255 bytes. Instead of implementing fragmentation, I modified the protocol to use a `short` for length. This requires an extra step in the state machine to parse the second byte, which is why the logic differs from the previous if-statement.

android-src/KV4PHT/app/src/main/java/com/vagell/kv4pht/radio/Protocol.java
dkaukov: Since we added a byte for the length, I shortened the delimiter to compensate. I also believe 0xDEADBEAF is distinct enough to minimize the risk of false positives in random data.

microcontroller-src/kv4p_ht_esp32_wroom_32/protocol.h
```diff
@@ -94,6 +96,13 @@ struct config {
 } __attribute__((__packed__));
 typedef struct config Config;
 
+// COMMAND_WINDOW_ACK parameters
+struct window_update {
```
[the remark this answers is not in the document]
dkaukov: This implements a basic HTTP/2 flow control mechanism: the sender decreases the window as it transmits data, and the receiver sends WINDOW_UPDATE frames to replenish it. This prevents buffer overflow and ensures efficient data transfer. I'll add a more detailed explanation to the docs after the merge.

android-src/KV4PHT/app/src/main/java/com/vagell/kv4pht/radio/Protocol.java
dkaukov: I will update after we merge.

microcontroller-src/kv4p_ht_esp32_wroom_32/buttons.h
dkaukov: I think this should be tracked as a separate ticket. My theory is that all input pins require stronger pull-ups/pull-downs than the internal ones. I don’t have PCB v2 to test, so someone with access will need to verify.

microcontroller-src/kv4p_ht_esp32_wroom_32/kv4p_ht_esp32_wroom_32.ino
dkaukov: On my modules, I've observed bounces during TX → RX transitions, which are even visible on the RGB LED. Initially, I thought it was a software bug, but after checking with an oscilloscope, I confirmed the transitions. That's why I think debouncing is necessary.

.github/workflows/arduino.yml
dkaukov: done

VanceVagell:
Merging this now since it generally works. I need to do further testing to determine if the stick PTT issue is from this PR or not, and also need to revert to older releases until the buffering issue goes away so we can pin down when it was introduced. Those two issue will block next release, but these reworks in this PR are in good shape, thanks Dmitry.

PR #257, opened 2025-04-15
Added i18n support, with zh_CN localization

NaiveTomcat:
As the user of this project, ranging all over the world, may not be fluent in English, here I present the refactorization of the Android application with i18n support, and also with zh_CN localization variant.

The main changes are below:
- Moved (nearly) all user-readable texts from layout XML and Java code to resource string file, making localization possible
- Added zh_CN translation of said string file, presenting localization for this locale.


android-src/KV4PHT/app/src/main/java/com/vagell/kv4pht/ui/MainActivity.java
```diff
@@ -837,8 +837,8 @@ private void showBeaconingOnSnackbar(int accuracy) {
             return;
         }
 
-        String accuracyStr = (accuracy == RadioAudioService.APRS_POSITION_EXACT) ? "exact" : "approx";
-        CharSequence snackbarMsg = "Beaconing your " + accuracyStr + " position on active frequency";
+        String accuracyStr = (accuracy == RadioAudioService.APRS_POSITION_EXACT) ? getString(R.string.exact) : getString(R.string.approx);
+        CharSequence snackbarMsg = getString(R.string.position_beacon_message_1) + accuracyStr + getString(R.string.position_beacon_message_2);
```
dkaukov, line 841: Consider using a formatted string resource with placeholders (e.g., '%s') instead of concatenating multiple string resources. This approach will provide better flexibility for languages with different word orders.

android-src/KV4PHT/app/src/main/res/values/strings.xml
```diff
@@ -7,40 +7,92 @@
...
+    <string name="advanced">Advanced</string>
+    <string name="bandwidth">Bandwidth</string>
+    <string name="min_2m_tx_frequency">Min 2m TX frequency</string>
+    <string name="max_2m_tx_frequency">Max 2m TX frequency</string>
+    <string name="min_70cm_tx_frequency">Min 70cm TX frequency</string>
+    <string name="max_70cm_tx_frequency">Max 70cm TX frequency</string>
+    <string name="mic_gain_boost">Mic gain boost</string>
+    <string name="to" translatable="false">to</string>
+    <string name="kv4p_aprs_message" translatable="false">KV4P</string>
+    <string name="message_body_example">The quick brown fox jumped over the lazy dog.</string>
```
VanceVagell, line 69: I think this is probably left over from an example/template and could be removed.

android-src/KV4PHT/app/src/main/java/com/vagell/kv4pht/ui/MainActivity.java
VanceVagell: I agree with @dkaukov , and there are only a handful of cases of this in the app where we have a concatenated string so it shouldn't be hard to fix. If you're available to submit another PR to fix this, please do so @NaiveTomcat , but I won't block this PR on it because it works as-is for the 2 currently implemented languages (en and cn).

VanceVagell:
Thank you so much for this! There are many kv4p HT users in China (and elsewhere) who I am sure will appreciate this improvement. A couple code review comments were left, if you have time to submit a small follow up PR please do. Appreciate the help on our project!

PR #262, opened 2025-04-18
v2.1b: VHF or UHF.  Support 80x50x20mm metal case.

SmittyHalibut:
The three primary changes are:
* Added a common mode current choke to the USB differential data pair.
  * We've had reports of USB disconnecting when transmitting. This will hopefully address that problem.
  * Power pins already had RF beads on them, but the differential pair needs a full CMCC.
* Support for either VHF or UHF modules.
  * Band is selected by toggling Do Not Populate (DNP) flags on various components (documented in schematic, see below.)
  * Several decoupling caps are Murata parts tuned for resonance at the operating band, to maximize decoupling effectiveness.  They also need to be DNP'd based on the band.
* Resize the PCB to 80x47mm, to fit an 80x50x20mm extruded aluminum case.  Example: https://www.aliexpress.us/item/3256806642271098.html
  * Also added front and back panel boards for that case.
  * The PTT buttons had to come in from the edge of the board to fit in the case.  So any 3D printed cases will need to extend in-side the board outline to reach the PTT buttons.

VHF/UHF Selection:
image

This design is entirely backward compatible with v2.0d, so it uses the same board identifier.


SmittyHalibut:
Some images of the board:
image

image

image

All that DNP wackiness results in the following.
## VHF:
image

image

## UHF:
image

image

VanceVagell:
Awesome improvements, looking forward to using this!

PR #271, opened 2025-04-30
Proposed approach to fix NPE's in future

dkaukov:
1) Make `threadPoolExecutor` non-null and final.  
2) In `MainActivity::applySettings()`, check `radioAudioService.isRadioConnected()` before proceeding.  
3) In `MainActivity::processAudioStream()`, throw an `IllegalStateException` if the radio isn’t connected. This won’t impact users, but it’ll help track how often USB gets "lost" during TX.


VanceVagell:
Thanks for the improvements. Since ThreadPoolExecutor is used in all the Activities, could you please apply this change there as well so it's consistent throughout the project?

VanceVagell:
Will monitor this build before releasing it, to see if that one corner case that was commented comes up (it may have been fixed in other improvements since it was written, we'll see). Thanks again!

Issue #237, opened 2025-02-22
[Android] Add more ways to tune/change frequency


norrinr:
Please add more ways to change the frequency beyond typing in the number. Maybe a slider bar/wheel or - / + with a step?

raff:
Also, I though "scan" would scan the band, not just cycle through the memorized frequencies (and I couldn't understand why "scan" wasn't doing anything until I added a couple of channel :)

Would be good to have a "scan band" command, that stops when it finds a signal.

VanceVagell:
@raff we're tracking the request to scan frequencies in #94 so keep an eye on that for future developments!

In terms of adding buttons to move up and down by small amounts, this is not currently planned for the project, since the use case for it is limited and a goal of the project is to focus on features most people need so the radio remains more approachable. The reality for most hams is that they either want to tune to a known simplex frequency to talk to buddies (walkie talkie style) or they want to program in nearby repeaters (or use our "Find nearby repeaters") function to tune to specific repeater frequencies.

There are certainly cases where moving up/down a little bit might be helpful, but it's niche enough that we wouldn't change the main UI to accomodate it. Thanks for the idea though!

PR #275, opened 2025-05-02
Add buttons to increase/decrease frequency

raff:
This seems to work. If you are interested I can go ahead and add a setting to specify the delta (now fixed at 100Khz).

See https://github.com/VanceVagell/kv4p-ht/issues/237

### DIFF
--- a/android-src/KV4PHT/app/src/main/java/com/vagell/kv4pht/radio/RadioAudioService.java
+++ b/android-src/KV4PHT/app/src/main/java/com/vagell/kv4pht/radio/RadioAudioService.java
@@ -136,7 +136,7 @@ public class RadioAudioService extends Service {
 
     // Callbacks to the Activity that started us
     @Setter
-    private @NonNull RadioAudioServiceCallbacks callbacks = NO_OP_CALLBACKS;
+    public @NonNull RadioAudioServiceCallbacks callbacks = NO_OP_CALLBACKS;
 
     // For transmitting audio to ESP32 / radio
     public static final int AUDIO_SAMPLE_RATE = 48000;
@@ -336,6 +336,10 @@ public class RadioAudioService extends Service {
         }
     }
 
+    public float getMinRadioFreq() {
+        return minRadioFreq;
+    }
+
     public void setMaxRadioFreq(float newMaxFreq) {
         maxRadioFreq = newMaxFreq;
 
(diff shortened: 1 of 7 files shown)


VanceVagell:
Very cool, thanks for looking into this. Just a heads up I'm traveling for most of the next week, so my review will be delayed.

VanceVagell:
Just pinging this PR so you know I haven't forgotten about it! Just a bit behind. I plan to checkout this PR and test it, but I could review this PR faster if you have a short video demoing it. Otherwise it will be some night or weekend soonish (apologies!)

raff:
I have a hard time getting a video since the controls don't work if the radio is not connected and I can't connect to AndroidStudio if I connect the radio (is there a way to run the simulator with the serial port connected) ?

here are a couple of screenshots. Latest version uses the bandwidth value (25k or 12.5k) as the increment value:

Screenshot_20250519_215332_kv4p HT
Screenshot_20250519_215215_kv4p HT

VanceVagell:
OK I had a chance to test this out, sorry for the delay. I need to recruit some more code reviewers familiar with Android UI. :)

This is pretty close, just a few changes:

* Please use the icons "Keyboard arrow up" and "Keyboard arrow down" instead of +/-, since we use + for "Add memory" and it could be confusing to have both on the screen with the same icon.
* There is a layout problem on wider screens where the up/down buttons are too close to the screen edge, see this example:

share_7917619435230389702

To fix this, please use 24dp margin from the left, and the right. And, ideally these buttons would be hidden if there's not enough space for them horizontally (very narrow screens). But I think that's probably a corner case, if it's not easy to do that.

* Please disable the up/down buttons when the frequency is already too close to the top/bottom of band to let the user change the frequency anymore in that direction. Use the full range of frequencies for this, not just ham band (e.g. use 134-174MHz for VHF).

I think that's it... will give it one more try after you make these edits. Thank you!

raff:
What font or symbols do you use for the icons ? And what exact characters do you want ? I can change them but I want to make sure I get the right ones.

I also got kind of busy, but I'll work on this in the next few days (I have some changes almost ready, but I need to test them)

VanceVagell:
You can find the icons here (search for the names I mentioned):

https://fonts.google.com/icons

They just need to recolored to match other icons in the app. You can see the others icons in res > drawable in the project, this is where the icon xml should go. IIRC I used right-click > New > Vector asset on that folder in Android Studio to import the raw icon SVG into an Android XML file, then I recolored it in the layout I think.

VanceVagell:
Closing this PR since it seems to have stalled, please reopen it if/when ready to be reviewed!

PR #281, opened 2025-05-08
Fix: Add spacing between to and callsign in APRS message view

raveendra11:


android-src/KV4PHT/app/src/main/res/layout/aprs_message.xml
```diff
@@ -29,7 +29,8 @@
             android:layout_height="wrap_content"
             android:text="@string/to"
             android:textColor="@android:color/darker_gray"
-            android:textSize="16sp" />
+            android:textSize="16sp" 
+            android:layout_marginRight="4dp" />
```
VanceVagell, line 33: Thanks for working on this! Instead of changing the layout, please edit the English string in the text resource to add a space. The space might not make sense in other languages.

raveendra11: Thanks for the suggestion! I've updated the English string resource to include a trailing space:

**<string name="to">to </string>**

Issue #282, opened 2025-05-08
Proposal: Use NVS for Centralized Board Hardware Configuration


dkaukov:
We're proposing to use NVS (`nvs_flash`) to store board-specific hardware configuration (`hwconfig`), such as pin mappings and enabled features. This would allow us to support multiple hardware variants with a single firmware binary, without cluttering the codebase with conditionals and quirks.

📄 **Comparison Matrix:**
| Approach                                    | Pros                                                                                                                                                                                                                                                                                                                                                                             | Cons                                                                                                                                                                                                                        |
| ------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| **Centralized `hwconfig` in NVS**           | - Single firmware binary for all boards<br>- Clean and structured config<br>- “Everything enabled” config covers most cases<br>- Scales easily as new boards appear<br>- NVS only needed for non-standard designs<br>- Breaks the dependency between hardware and firmware release cycles — tweaks don't need new firmware<br>- @SmittyHalibut  is happy to preflash 2.x boards with NVS | - Slightly more setup for custom boards<br>- Recovery from corrupted/missing NVS could confuse users<br>- Requires some knowledge/tools to manage NVS                                                                       |
| **GPIO-based detection + scattered tweaks** | - Also a single firmware binary<br>- Works automatically for known designs<br>- Doesn’t require users to understand NVS — just connect resistors to strap pins according to table                                                                                                                                                                                                | - Firmware gets cluttered with board-specific quirks<br>- Harder to scale cleanly with new board variants<br>- Requires firmware updates for every new layout |


**Benefits:**

* Clean separation of firmware and hardware config
* One firmware binary for all boards
* Easier testing: “everything enabled” config works nearly everywhere
* Simplifies support for future boards without firmware changes
* No need to wait for a firmware release just to support a new layout

**Considerations:**

* NVS is only needed for non-standard/custom boards
* Adds a bit of complexity for people deviating from the standard layout
* Recovery from corrupted/missing NVS needs to be understandable
* Some concerns about DIY usability and tooling (Android app, online flasher?)

SmittyHalibut:
I 100% agree this is the right solution going forward.

I disagree with a few details provided in the summary:

[...]

PR #283, opened 2025-05-08
Proposal: Use NVS for Centralized Board Hardware Configuration

dkaukov:
Use NVS (nvs_flash) to store board-specific hardware configuration (hwconfig), such as pin mappings and enabled features. This would allow us to support multiple hardware variants with a single firmware binary, without cluttering the codebase with conditionals and quirks.

Fixes: https://github.com/VanceVagell/kv4p-ht/issues/282


android-src/KV4PHT/app/build.gradle
```diff
@@ -70,4 +70,6 @@ dependencies {
     implementation 'org.projectlombok:lombok:1.18.30'
     annotationProcessor 'org.projectlombok:lombok:1.18.30'
     implementation 'io.github.jaredmdobson:concentus:1.0.1'
+    implementation 'io.github.dkaukov:esp32-flash-lib:1.1.15'
+    implementation 'dev.bright.slf4android:slf4android:0.1.8'
```
VanceVagell, line 74: Please add license text to other-licenses.txt for any libraries used (even those you own/created).

VanceVagell:
This is fantastic, thanks Dmitry! One small change requested.

android-src/KV4PHT/app/build.gradle
dkaukov: done

VanceVagell:
OK I started taking a look at this, but there's a LOT going on in this PR. I'd like to focus my attention on the Android UI questions you had. Could you please list what those questions are (i.e. which screens, and what you're trying to figure out)?

Once you've got the UI fixed up, I'll probably just test this functionally since I'm less familiar with some of what you're doing here. Adding @wsciaroni in case he has any feedback on the firmware changes (but I understand if you don't have the time/bandwidth!)

android-src/KV4PHT/app/src/main/res/raw/firmware_v14.bin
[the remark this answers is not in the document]
dkaukov: Got it. I included the binary just to test the flasher and NVS persistence. Not a big deal — we can drop it if it doesn’t fit the usual flow. Just figured I’d ask if it’s cool to leave it for now?

dkaukov:
> Once you've got the UI fixed up, I'll probably just test this functionally since I'm less familiar with some of what you're doing here. Adding @wsciaroni in case he has any feedback on the firmware changes (but I understand if you don't have the time/bandwidth!)

Hi @VanceVagell — thanks for taking a look! Totally fair, there’s a lot going on in this PR.

The Android UI is in kind of a “working” state right now, but still needs proper wiring. Here's where I'm at:

* The settings screen now shows a UHF/VHF toggle, but it’s currently ignored.
* The H/L (high/low power) setting isn’t exposed in the UI — it’s just a property in `RadioService.java` that gets sent to the ESP32.
* There *is* a feature toggle for this in `RadioService`, but:

  * It’s not hooked into the UI.
  * It depends on the ESP32 being connected — which means the toggle isn't even available if the radio isn’t connected.

Next step is probably to extend the settings schema and update the relevant layout XMLs to expose the H/L setting. I’m happy to do that — just need a pointer to where those changes should go, since I’m not too familiar with the XML/layout side of things.

Also: any thoughts on what we should do when the ESP32 isn’t connected and the toggle can’t be determined? Should we block access to the settings screen, or just show a reduced version?

Appreciate the help!

VanceVagell:
I realized I didn't answer your question about what to do before the kv4p
HT is plugged in. Let's disable any settings (not hide them) that are
version specific but leave the general settings always available to
change.  Same thing for when a user plugs in a kv4p HT that doesn't support
every setting, just disable those rather than remove them.

This way people aren't confused about settings disappearing. If they're
confused about the disabled settings it should be pretty straightforward to
figure out what's going on.

dkaukov:
Hi @VanceVagell — thanks, that was really helpful!

I think it’s ready to test now. Quick summary of the UI changes:

1. The band (VHF/UHF) is now shown on the main screen.
2. A new *Radio Power* setting has been added to the settings screen.
3. The RF power setting is always visible, but disabled when the connected device doesn’t support it — in those cases, it defaults to showing “High” as a placeholder.
4. I did a bit of refactoring to satisfy SonarQube — mainly removing `static`s, introducing proper singletons, and simplifying duplicated logic.

VanceVagell:
Just a quick note to confirm I haven't forgotten this PR. I plan to review it within the week, thanks for the hard work on it!

SmittyHalibut:
I'm WAY behind on my reviews.  Starting now. Finally.

> Also: any thoughts on what to do when the ESP32 isn’t connected and the toggle can’t be determined? Should we block access to the settings screen, or just show a reduced version?

As a user, not an app developer, it would make more sense to me for the options to be there, but greyed out if I can't change them.

I think the same applies to settings that your hardware doesn't support, like v1 boards and the High/Low setting.

Aaaand, it looks like Vance already said the same thing.  *whew!*  Glad to know I'm not totally wrong. ;-)

> Just a quick note to confirm I haven't forgotten this PR. I plan to review it within the week, thanks for the hard work on it!

Not just you.  :-)   Thank you @dkaukov for your patience, and ALL THIS WORK.

Alright, caught up on comments, diving into firmware.  I'm not going to try looking at the Android code, only the ESP32 code.

microcontroller-src/kv4p_ht_esp32_wroom_32/board.h
```diff
@@ -0,0 +1,131 @@
+/*
+KV4P-HT (see http://kv4p.com)
+Copyright (C) 2024 Vance Vagell
```
SmittyHalibut, line 3: 2025.  :-)

microcontroller-src/kv4p_ht_esp32_wroom_32/board.h
```diff
@@ -0,0 +1,131 @@
...
+  hw.rfModuleType = (RfModuleType) prefs.getUChar("RF_MODULE_TYPE", DEFAULT_RF_MODULE_TYPE);
+  prefs.end();
+}
+
+void saveHardwareConfig() {
+  prefs.begin("hwconfig", false); // read-write mode
+  prefs.putBool("HWCONFIG",        true);
+  prefs.putChar("RXD2_PIN",        hw.pins.rxd2Pin);
+  prefs.putChar("TXD2_PIN",        hw.pins.txd2Pin);
+  prefs.putChar("DAC_PIN",         hw.pins.dacPin);
```
SmittyHalibut: I assume Preferences optimizes writes to local flash?  I twitch a little bit by writing every parameter even if it hasn't changed.  But the .begin() and .end() give Preferences the opportunity to do the right thing; I'm just not sure it does.

Do you know either way?

microcontroller-src/kv4p_ht_esp32_wroom_32/board.h
```diff
@@ -0,0 +1,131 @@
...
+  prefs.putChar("HL_PIN",          hw.pins.hlPin);
+  prefs.putChar("ADC_ATTEN",       hw.adcAttenuation);
+  prefs.putString("ADC_BIAS",      String(hw.adcBias, 6));
+  prefs.putBytes("STOPPED_COLOR",  &hw.stoppedColor, sizeof(RGBColor));
+  prefs.putUChar("VOLUME",         hw.volume);
+  prefs.putUChar("RF_MODULE_TYPE", hw.rfModuleType);
+  prefs.end();
+}
+
+hw_ver_t get_hardware_version() {
```
SmittyHalibut, line 99: Why is `get_hardware_version` not `inline` but `boardSetup` is?  I'm still figuring out this whole header-only thing, so help educate me.  I'm still used to only inline-ing small things to optimize performance.

Also: `getHardwareVersion()`  camelCase.

microcontroller-src/kv4p_ht_esp32_wroom_32/globals.h
```diff
@@ -84,5 +73,64 @@ bool squelched = false;
...
+    .pixelsPin = DEFAULT_PIXELS_PIN,
+    .hlPin = DEFAULT_HL_PIN,
+  },
+  .features = {
+    .hasHL = (DEFAULT_HL_PIN != -1),
+    .hasPhysPTT = (DEFAULT_PHYS_PTT_PIN1 != -1 || DEFAULT_PHYS_PTT_PIN2 != -1)
+  },
+  .adcBias = DEFAULT_ADC_BIAS_VOLTAGE,
+  .adcAttenuation = DEFAULT_ADC_ATTENUATION,
+  .volume = 8,
```
SmittyHalibut, line 133: `DEFAULT_VOLUME`

microcontroller-src/kv4p_ht_esp32_wroom_32/globals.h
```diff
@@ -84,5 +73,64 @@ bool squelched = false;
...
+  .stoppedColor = {0, 32, 0},
```
SmittyHalibut, line 134: Make a `DEFAULT_STOPPED_COLOR` above and use it here.  These two magic constants stick out like sore thumbs when everything else is so nicely #define'd.  :-)

microcontroller-src/kv4p_ht_esp32_wroom_32/kv4p_ht_esp32_wroom_32.ino
```diff
@@ -98,54 +87,42 @@ void setup() {
...
-      COLOR_HW_VER = {32, 0, 0};
-      sqPin = SQ_PIN_HW2;
-      break;
-    case HW_VER_V2_0D:
-      COLOR_HW_VER = {0, 0, 32};
-      sqPin = SQ_PIN_HW2;
-      break;
-    default:
-      // Unknown version detected. Indicate this some way?
-      COLOR_HW_VER = {16, 16, 16};
```
SmittyHalibut, line 117: Here's a good `DEFAULT_STOPPED_COLOR` candidate.

microcontroller-src/kv4p_ht_esp32_wroom_32/kv4p_ht_esp32_wroom_32.ino
```diff
@@ -98,54 +87,42 @@ void setup() {
...
-  pinMode(PTT_PIN, OUTPUT);
-  digitalWrite(PTT_PIN, HIGH);  // Rx
+  pinMode(hw.pins.pdPin, OUTPUT);
+  digitalWrite(hw.pins.pdPin, HIGH);  // Power on
+  pinMode(hw.pins.sqPin, INPUT);
+  pinMode(hw.pins.pttPin, OUTPUT);
+  digitalWrite(hw.pins.pttPin, HIGH);  // Rx
+  if (hw.features.hasHL) {
+    pinMode(hw.pins.hlPin, OUTPUT);
+    digitalWrite(hw.pins.hlPin, LOW);  // High power
```
SmittyHalibut: /me removes these two lines from his PR.  :-)

microcontroller-src/kv4p_ht_esp32_wroom_32/globals.h
```diff
@@ -84,5 +73,64 @@ bool squelched = false;
...
+    .dacPin = DEFAULT_DAC_PIN,
+    .adcPin = DEFAULT_ADC_PIN,
+    .pttPin = DEFAULT_PTT_PIN,
+    .pdPin = DEFAULT_PD_PIN,
+    .pttPhys1 = DEFAULT_PHYS_PTT_PIN1,
+    .pttPhys2 = DEFAULT_PHYS_PTT_PIN2,
+    .ledPin = DEFAULT_LED_PIN,
+    .pixelsPin = DEFAULT_PIXELS_PIN,
+    .hlPin = DEFAULT_HL_PIN,
+  },
```
SmittyHalibut: Some names are defined by the application function: ptt, pd (power down), ptt.  But others are defined by the microcontroller function: rxd2, adc, dac.

This is absolutely not a release blocker, but in general, I think we should make them all application based.  ie: rename rxd2 to rxFromRadio, txd2 to txToRadio, adc to rxAudio, dac to txAudio, etc.

Looking over them, I think those are the only ones.  The serial port and audio pins.  Again, not a release blocker, just expressing a style thing I noticed.

microcontroller-src/kv4p_ht_esp32_wroom_32/led.h
```diff
@@ -20,30 +20,19 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
-  uint8_t red;
-  uint8_t green;
-  uint8_t blue;
-};
 const RGBColor COLOR_STOPPED = {0, 0, 0};
 const RGBColor COLOR_RX_SQL_CLOSED = {0, 0, 32};
 const RGBColor COLOR_RX_SQL_OPEN = {0, 32, 0};
 const RGBColor COLOR_TX = {16, 16, 0};
 const RGBColor COLOR_BLACK = {0, 0, 0};
-RGBColor COLOR_HW_VER = COLOR_BLACK;
```
SmittyHalibut, line 40: Oh, apparently this is the `DEFAULT_STOPPED_COLOR`.  Whatever.  Just don't define a magic constant in the middle of the code.  :-D

microcontroller-src/kv4p_ht_esp32_wroom_32/protocol.h
```diff
@@ -91,8 +95,8 @@ typedef struct filters Filters;
 #define FILTER_LOW  (1 << 2)
 
 // COMMAND_HOST_CONFIG parameters
-struct config {
-  uint8_t radioType; 
+struct config { 
+  bool isHigh;   
 } __attribute__((__packed__));
 typedef struct config Config;
```
SmittyHalibut, line 101: Is `isHigh` a typo here?  I'm guessing it's an over-reaching copy-paste from creating `hlState` below.

Maybe `rfModuleType`?

microcontroller-src/kv4p_ht_esp32_wroom_32/rxAudio.h
```diff
@@ -78,15 +78,11 @@ DCOffsetRemover dcOffsetRemover(DECAY_TIME, AUDIO_SAMPLE_RATE);
...
-  dac_output_voltage(DAC_CHANNEL_2, (255.0 / 3.3) * ADC_BIAS_VOLTAGE);
+  dac_output_voltage(DAC_CHANNEL_2, (255.0 / 3.3) * hw.adcBias);
 } 
 
 inline void setUpADCAttenuator() {
-  if (hardware_version == HW_VER_V2_0C) { // v2.0c has a lower input ADC range
-    adc1_config_channel_atten(I2S_ADC_CHANNEL, ADC_ATTENUATION_v20C);
-  } else {
-    adc1_config_channel_atten(I2S_ADC_CHANNEL, ADC_ATTENUATION);
-  }
```
SmittyHalibut, line 89: Aaahhhhh!  Ok, NOW I understand what you mean by "doing away with all the conditionals everywhere."  Yes yes yes.

pcb/scripts/nvs_partition_gen.py
```diff
@@ -0,0 +1,1318 @@
+#!/usr/bin/env python3
```
SmittyHalibut, line 1: ngl, I did NOT review this.  :-)

This generates the binary blob that `Preferences.h` uses?  Is that what this does?

microcontroller-src/kv4p_ht_esp32_wroom_32/board.h
```diff
@@ -0,0 +1,131 @@
...
+  return exists;
+} 
+
+void loadHardwareConfig() {
+  prefs.begin("hwconfig", true); // read-only
+  hw.pins.rxd2Pin   = prefs.getChar("RXD2_PIN",     DEFAULT_RXD2_PIN);
+  hw.pins.txd2Pin   = prefs.getChar("TXD2_PIN",     DEFAULT_TXD2_PIN);
+  hw.pins.dacPin    = prefs.getChar("DAC_PIN",      DEFAULT_DAC_PIN);
+  hw.pins.adcPin    = prefs.getChar("ADC_PIN",      DEFAULT_ADC_PIN);
+  hw.pins.pttPin    = prefs.getChar("PTT_PIN",      DEFAULT_PTT_PIN);
```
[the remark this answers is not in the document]
dkaukov: Done

pcb/v1.5/board-config.bin
SmittyHalibut, line 1: Should the binary files be built from the .csv as part of the release process?

pcb/v1.8c (30pin)/board-config.csv
```diff
@@ -0,0 +1,18 @@
+key,type,encoding,value
```
SmittyHalibut: I noticed the UHF versions are called `-uhf`, but the VHF aren't `-vhf`.  I'd prefer these be symmetric.

SmittyHalibut (commented):
Over all, I REALLY like this change.  SO MUCH cleaner.

I review the microcontroller stuff; I didn't review the Android stuff, since I'm worse than useless there.

microcontroller-src/kv4p_ht_esp32_wroom_32/board.h
dkaukov: ;)

microcontroller-src/kv4p_ht_esp32_wroom_32/globals.h
dkaukov: done

microcontroller-src/kv4p_ht_esp32_wroom_32/protocol.h
dkaukov: Nah. Android now sends initial H/L state and not sending ModuleType as it is configured in NVS. Hence protocol change.

pcb/scripts/nvs_partition_gen.py
dkaukov: This script is originally part of the ESP-IDF project (see: https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-reference/storage/nvs_partition_gen.html), but it’s been extracted and committed here for convenience—no need to install ESP-IDF.

pcb/v1.5/board-config.bin
SmittyHalibut: Fair point.  Ok.

SmittyHalibut:
I approve the current PR, but it looks like it needs some merge conflict resolutions before a merge can happen.  Also, ya know, Vance reviewing the Android app part.

But it looks good to me.  👍🏼

VanceVagell:
OK I've now also tested with a 1.8x PCB and a UHF radio module. I see a few issues we need to fix / clarify for me how these are meant to work:

- The "Power" drop-down in settings only ever has the "High" option in it (even if I attach a v2.0x kv4p HT). Is there something else I need to do to get this to include the "Low" option (e.g. change something before flashing the firmware)? If the attached module doesn't support changing this, this option should be present but disabled (i.e. the drop-down is grayed out and can't be tapped, but does say "High" since that's the default fixed power in that case).
- There's no longer a radio type drop-down, so when I attach the v1.8x board with a UHF module on it, it seems there's no way for me to tell the app that. Is this PR assuming this must be defined in the firmware? If so, it should instead default to letting the user choose it, and only use the firmware-provided radio module type if it's explicitly been set (i.e. guaranteed to be true).

Otherwise this PR seems to function well. Please fix/comment on the above and we can move it forward. Thanks Dmitry!

dkaukov:
I guess as a short-term solution, we could add a `HwConfig` section to the website:
[https://github.com/dkaukov/kv4p-ht/compare/nvs...dkaukov\:kv4p-ht\:nvs-website?expand=1](https://github.com/dkaukov/kv4p-ht/compare/nvs...dkaukov:kv4p-ht:nvs-website?expand=1)

But again, this is only needed for **v1.x UHF** users — which I believe is a small percentage of the user base.
For most users, the config should be transparent and just work out of the box.

dkaukov:
Thanks! Let me know if I can help with anything.

Also, I put together an example of how we could use the web flasher to flash NVS — for folks who don’t want to mess with `esptool`:
[https://github.com/dkaukov/kv4p-ht/compare/nvs...dkaukov\:kv4p-ht\:nvs-website?expand=1](https://github.com/dkaukov/kv4p-ht/compare/nvs...dkaukov:kv4p-ht:nvs-website?expand=1)

Should make things a bit smoother for DIY users.

dkaukov:
Tnx!

1. The missing default for the power drop-down should be fixed now:
   [https://github.com/VanceVagell/kv4p-ht/pull/283/commits/77a8872c74b5894a98d348c5bc5093182d3cd589](https://github.com/VanceVagell/kv4p-ht/pull/283/commits/77a8872c74b5894a98d348c5bc5093182d3cd589)

2. Re: TX — I’ll test tonight with a power meter and see if it's behaving normally here. It’s possible something unrelated is going on (e.g. TX offset not getting loaded?), but I’ll dig in and let you know what I find.

dkaukov:
Hi @VanceVagell — just checked, and I’m not seeing the TX issue on my side. Everything seems to be working fine here.

VanceVagell:
OK thanks, let's merge this one then! I'll retest before releasing a new version.

PR #293, opened 2025-06-16
Firmware code to test hardware. Helps with assembly.

SmittyHalibut:
I've historically maintained a separate program that exercises the hardware.  I'll flash this onto boards during assembly to confirm various things: buttons work, LEDs work, serial communication with the SA818 works, etc. I then have to re-flash the boards with the production firmware before shipping to the customer.

Instead of flashing twice, I've integrated this code into the main firmware.  Now the hardware test code is triggered by holding both PTT buttons when the board resets.  Either by pressing the Reset button, or by letting the watchdog timer expire.  That is: Connecting it to a power source (eg: battery, or computer) that's NOT running the KV4P App, and holding both PTT buttons for several seconds (at least a watch-dog timeout), will put it into hardware test mode.

Aaaaand, apparently, there are a bajillion "Merge" commits that don't actually do anything. I'm not sure how to get rid of those.

Also: Porting my test code to use `globals.h` involved making a few changes to the main code. Mostly around not defining code or variables in .h files.  You can DECLARE them, but not DEFINE them.  (This is a major difference between C and Java.)  So I had to move some things around, but it should all still be there.


dkaukov (changes requested):
Hi @SmittyHalibut, nice work!

Couple quick thoughts:

1. Would be great to land [https://github.com/VanceVagell/kv4p-ht/pull/283](https://github.com/VanceVagell/kv4p-ht/pull/283) first. It touches some of the same files, so merging this after should make life easier.

2. Pretty sure you’re flashing the full firmware before shipping mostly because the in-app flasher is slow and unreliable. That makes sense. But #283 adds a much faster, verified flasher — so maybe in the future we can skip pre-flashing the full firmware, just flash a tiny self-test, test, and let the app flash the real thing later.

3. The test uses PTT to enter test mode, but that’s a bit of a chicken-and-egg situation if PTT isn’t working. Might be worth adding a GPIO strap or protocol command to trigger it — that way, the whole process could be scripted end-to-end.

4. We’ve been doing [header-only](https://en.wikipedia.org/wiki/Header-only) so far. Doesn’t seem like we need a separate `.cpp` just for the test code — any reason not to keep it inline?

microcontroller-src/kv4p_ht_esp32_wroom_32/hardware-test.cpp
```diff
@@ -0,0 +1,238 @@
...
+#endif
+const bool IF_SERIAL=USE_SERIAL;
+
+void print_buttons() {
+    static bool PTT_Left, PTT_Right, Program;
+    bool tmp;
+
+    tmp = digitalRead(PIN_BUTTON_PTT_LEFT);
+    if (PTT_Left != tmp) {
+        IF_SERIAL && Serial.print("PTT Left: ");
```
dkaukov, line 25: We already have the _LOGI() macro — maybe it’s better to use that instead of Serial.print()? It also supports format strings, which might simplify things a bit.

microcontroller-src/kv4p_ht_esp32_wroom_32/hardware-test.cpp
```diff
@@ -0,0 +1,238 @@
...
+    }
+    return 0;
+}
+
+float line(uint16_t reading, float start_v, uint16_t start_r, float end_v, uint16_t end_r) {
+    float v = (end_v - start_v) * float(reading - start_r)/(end_r - start_r) + start_v;
+    return float(int(v*100.0))/100.0;
+}
+
+float adc_map(uint16_t reading) {
```
dkaukov, line 61: We're kind of using camelCase across the project — could you update this to match?

microcontroller-src/kv4p_ht_esp32_wroom_32/hardware-test.cpp
```diff
@@ -0,0 +1,238 @@
...
+        // 0 = 0.15, 2.7v = 3200
+        return line(reading, 0.15, 0, 2.70, 3200);
+    } else {
+        // 3.15v = 4095
+        return line(reading, 2.7, 3200, 3.15, 4095);
+    }
+}
+
+uint32_t last_time;
+bool PTT_STATE;
```
dkaukov, line 75: We’ve been using camelCase for variables across the project — only preprocessor macros are in ALL_CAPS.

SmittyHalibut:
Why won't github let me Reply to comments?  *grump*

>     1. Would be great to land [Proposal: Use NVS for Centralized Board Hardware Configuration #283](https://github.com/VanceVagell/kv4p-ht/pull/283) first. It touches some of the same files, so merging this after should make life easier.

CRIKEY, you're still waiting on me for that.  I am SO SORRY.  Yes, I'll go review that. Once it's merged into `main`, I'll rebase and re-submit this PR.  You get first dibs on `main.`

>     2. Pretty sure you’re flashing the full firmware before shipping mostly because the in-app flasher is slow and unreliable. That makes sense. But [Proposal: Use NVS for Centralized Board Hardware Configuration #283](https://github.com/VanceVagell/kv4p-ht/pull/283) adds a much faster, verified flasher — so maybe in the future we can skip pre-flashing the full firmware, just flash a tiny self-test, test, and let the app flash the real thing later.

I haven't looked at what all has changed in #283 yet (see above), so I might be talking out by stinky-hole here.  BUT.  Does the non-volatile memory mod improve the communication between the firmware and the app?

It was more of an issue that the Android app was having a hard time flashing, or detecting that a flash was needed, when it wasn't getting what it expected from the firmware.  My test firmware was outputting useful test data on the serial port and the app didn't know what to make of it.

@VanceVagell and I went back and forth a couple times trying to make the app more resilient to serial data it didn't expect, but we kept getting reports from users that it just wouldn't work.  They'd have to flash from the web browser (or from PlatformIO in my case; I saw the problem too) before the app would settle down.

So, I decided not to ship my boards with my test hardware at all.  I've been using it here to test the boards.  Found a NeoPixel with a dead Blue LED, and a SHORTED OUT SMA SOCKET (wtf??) today; glad I tested things first.   ANYhoo, I'll test the board, solder the SA818 (don't want to waste an SA818 on a board I won't be shipping), test it again to include testing the SA818, then re-flash the final firmware.

>     3. The test uses PTT to enter test mode, but that’s a bit of a chicken-and-egg situation if PTT isn’t working. Might be worth adding a GPIO strap or protocol command to trigger it — that way, the whole process could be scripted end-to-end.

Ooo, good call.  I hadn't thought about that.  Though, I guess if you're unable to put it into hardware test mode ... I mean... that's a pretty clear data point.  ;-)

I'll add holding a GPIO pin low as a way to get it into test mode.  One of the GPIO pins on the SAO or GPIO headers.

>     4. We’ve been doing [header-only](https://en.wikipedia.org/wiki/Header-only) so far. Doesn’t seem like we need a separate `.cpp` just for the test code — any reason not to keep it inline?

There were two functions that were not inline, the two that ended up in the `.cpp`.  And you can't define a global variable in a header that gets included more than once, even with `#pragma once`. At least, I was running into "already defined" errors until I made these changes.  Referencing `extern` https://learn.microsoft.com/en-us/cpp/cpp/extern-cpp?view=msvc-170

If you know how to do it correctly, please educate me. I do NOT claim to be an expert on "header-only" development.

I also don't much see the point; `inline` everywhere makes your code grow really quickly.  But that's a style argument that I won't press on at all.  I do agree that declaring and defining in one place makes way more sense, but unless we switch to Java-on-ESP32, or CircuitPython, we're stuck with C/C++.

But like I said, if we are doing header-only, and there's a way to make it work to include those headers more than once, I'll adjust. Consistency in coding style is way more important.

microcontroller-src/kv4p_ht_esp32_wroom_32/hardware-test.cpp
```diff
@@ -0,0 +1,238 @@
...
+
+// Previously defined in globals.h
+#define PIN_RADIO_PTT PTT_PIN
+#define PIN_RADIO_PD PD_PIN
+#define PIN_RADIO_SQUELCH SQ_PIN_HW2
+#define PIN_RX_FROM_RADIO RXD2_PIN
+#define PIN_TX_TO_RADIO TXD2_PIN
+
+#ifndef USE_SERIAL
+#define USE_SERIAL true
```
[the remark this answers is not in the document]
SmittyHalibut: It was something I did when the serial output from my test code was preventing the app from recognizing that it needed to flash the firmware.  Given the rest of the changes here, I can remove the IF_SERIAL/USE_SERIAL stuff.

microcontroller-src/kv4p_ht_esp32_wroom_32/hardware-test.cpp
SmittyHalibut: Unless you want me to avoid `Serial.print()` by itself with no conditional logic, I'll just remove the conditional.

microcontroller-src/kv4p_ht_esp32_wroom_32/hardware-test.cpp
```diff
@@ -0,0 +1,238 @@
...
+bool PTT_STATE;
+bool COR_STATE;
+
+// Set this as appropriate.  ESP32 is 12 bit.
+const uint8_t ADC_BITS = 12;
+const uint16_t ADC_MAX = (1<<ADC_BITS) - 1;
+
+bool HAS_PTT_RELEASED;
+
+void test_loop(void);
```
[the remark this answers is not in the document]
SmittyHalibut: Comment unclear.

`HAS_PTT_RELEASED` is needed because it's used in both `test_setup()` and `test_loop()`.

`test_loop();` is needed because I defined it after it's called in `test_setup()`.  I didn't declare `test_loop();` in `hardware-test.h` because no one should ever be calling it externally, its only scoped to this file.

Are you requesting me to define `test_loop()` above `test_setup()`?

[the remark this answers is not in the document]
SmittyHalibut: When `test_loop()` is removed:

```
Compiling .pio/build/esp32dev-release/src/kv4p_ht_esp32_wroom_32.ino.cpp.o
kv4p_ht_esp32_wroom_32/hardware-test.cpp: In function 'void test_setup()':
kv4p_ht_esp32_wroom_32/hardware-test.cpp:158:18: error: 'test_loop' was not declared in this scope
     while (true) test_loop();
                  ^~~~~~~~~
kv4p_ht_esp32_wroom_32/hardware-test.cpp:158:18: note: suggested alternative: 'test_setup'
     while (true) test_loop();
                  ^~~~~~~~~
                  test_setup
*** [.pio/build/esp32dev-release/src/hardware-test.cpp.o] Error 1
```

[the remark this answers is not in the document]
SmittyHalibut: I moved `testLoop()` (new name!) above `testSetup()`, since all the other local functions were up there too.  Hashtag Consistency.

microcontroller-src/kv4p_ht_esp32_wroom_32/hardware-test.cpp
SmittyHalibut: Done.

VanceVagell:
Hey Smitty, #283 is now merged if you'd like to update this one / wrap it up. Feel fee to merge this one youself after that, it looks good from my pespective.

VanceVagell (approved):

VanceVagell:
Hey Smitth, PR #301 is also now merged if you'd like to get this PR in.

SmittyHalibut:
Merging is ... problematic.  My important changes are mostly in unrelated code, so I'm going to withdraw this PR entirely and re-merge my changes in a different branch and submit another PR.

PR #302, opened 2025-07-14
HW config section on main website

dkaukov:


VanceVagell:
Thanks Dmitry. Generally this looks good, I'm just going to hold on merging it until I have some time to follow it up with some additional website edits before I push these changes to the website host. Having the PR reminds me this needs to happen, I know it's not ideal ;)

VanceVagell:
Just took another look at this -- does this flash the full firmware, or just the NVS contents? I believe the latter, but want to be sure.

If it's just NVS, I think only a subset of people will need access to this tool, so I'm likely to tweak this after I merge it so people don't get confused. I never liked the flashing tool hidden at the bottom of Quick Start, so I do like your idea of a new section. I might move all flashing-related stuff there, and just tweak some of the language.

Please clarify the current functionality of this, and we'll get it merged and ready for prime time.

Issue #300, opened 2025-07-12
[Android] update target API due to Google policy change


VanceVagell:
Within a few weeks the Play store won't accept apps with support for old Android APIs. Here's the details:

Starting August 31, 2025, new apps and app update submissions must target Android 15 (API level 35). Existing apps must target Android 14 (API level 34) or higher to remain available to new users on devices running Android OS higher than your app’s target API level. If you need more time, you can request an extension in Play Console to extend your deadline to November 1, 2025. See more details in Target API level requirements. To save time and effort, you can use the new SDK Upgrade Assistant in Android Studio to upgrade your app to the new Android SDK.

PR #303, opened 2025-07-16
[Android] update target API due to Google policy change

dkaukov:
Fixes https://github.com/VanceVagell/kv4p-ht/issues/300


dkaukov:
Fixes https://github.com/VanceVagell/kv4p-ht/issues/300

PR #304, opened 2025-07-21
Extracted updateFrequencyLimitsForBand() for use after settings reload

dkaukov:
Extracted updateFrequencyLimitsForBand() from setRadioType() so it can be called externally — like after reloading TX frequency settings. This makes sure txAllowed and band limits are correctly recalculated when settings change.

### DIFF
--- a/android-src/KV4PHT/app/src/main/java/com/vagell/kv4pht/radio/RadioAudioService.java
+++ b/android-src/KV4PHT/app/src/main/java/com/vagell/kv4pht/radio/RadioAudioService.java
@@ -693,26 +693,32 @@ public class RadioAudioService extends Service implements PacketHandler {
      */
     public void setRadioType(RadioModuleType radioType) {
         callbacks.setRadioType(radioType);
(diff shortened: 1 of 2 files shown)


android-src/KV4PHT/app/src/main/java/com/vagell/kv4pht/radio/RadioAudioService.java
```diff
@@ -693,26 +693,32 @@ public void onRunError(Exception e) {
      */
     public void setRadioType(RadioModuleType radioType) {
         callbacks.setRadioType(radioType);
-        if (!this.radioType.equals(radioType)) {
+        if (!Objects.equals(this.radioType, radioType)) {
```
VanceVagell, line 696: What is "Objects"? It's the only time it appears in this file.

VanceVagell: Nevermind, I see this is just a utility method on this Objects object.

Issue #299, opened 2025-07-08
NeoPixel glitches on 2.0c board when powered from laptop


dkaukov:
**What was found**:

* The NeoPixel data line has a 100 Ω resistor and a 2.2 nF cap (C13) to ground — meant as a low-pass filter.
* That filter cuts off around 723 kHz, which is too close to the NeoPixel’s 800 kHz signal.
* On top of that, ESP32 sends 3.3 V data, which is borderline.
* When plugged into a laptop, the USB voltage is a bit higher (5.1 V), which raises the NeoPixel’s threshold for "HIGH" — and now the signal sometimes doesn’t register.

**Fix**:
Removed C13 → NeoPixel works fine again even from the laptop.

I would suggest reducing C12/C13 to 100–220 pF — it’ll still help with high-frequency harmonics but won’t distort the NeoPixel timing.

SmittyHalibut:
Agreed.  100pF is a good choice.  This was on 2.0c?  I think I found and already fixed this on v2.1 actually; I ended up using a different neopixel chip that was much less tolerant of this problem.

SmittyHalibut:
Removed the filter entirely.  It's really not needed.

v2.0e PR incoming.  It mentions this ticket, so it should be linked soon. (I haven't submitted it yet, so I don't have a PR number.)

Resolving.

Issue #277, opened 2025-05-05
v2.0d PCB overlapping components


frank-ydf:
Hi, I found overlapped components on v2.0d PCB.
L3 and C27, L2 and C29, C31 and L1 are problematic if the PCB will be sent at JLCPCB for production.

SmittyHalibut:
I've ordered v2 boards no problems, they can manufacture it that close.

Having said that, I'm releasing a v2.0e board now that has them not quite so close.  I don't have a PR number yet because I'm still writing it.  But it will show up in this Issue because I've mentioned this one.

Resolving.

PR #307, opened 2025-07-30
PCB v2.0e: back-port some improvements from v2.1.

SmittyHalibut:
* Broke it out into its own versioned directory, instead of everything being in `KiCAD/`
* Added UHF support:
  * Added an outline for the MiniCircuit filter
  * Added all the 0R jumpers to DNP-Select between VHF and UHF.  All relevant components have a filed `VHF/UHF` that says which build to include or not include them for.
  * Made the 2.2nF tuned bypass caps dual-parts too, with the 220pF version for UHF.
* The complete rework to the filter section will address #277 too.
* Added a common mode current choke to the USB lines.
  * This involved moving the serial chip quite a bit.  But it turns out, that cleaned up a bunch of other stuff, so I'm happy with the change.
* Removed the ill-conceived filter capacitor from the NeoPixel data line. #299 

Otherwise, all the UI elements are in the same places: buttons, LEDs, connectors, etc.  It should be compatible with all v2.0 accessories (read: 3D printed cases.)

### DIFF
--- a/pcb/v2.0e/kv4p-ht/kv4p-ht.kicad_pcb
+++ b/pcb/v2.0e/kv4p-ht/kv4p-ht.kicad_pcb
@@ -809,7 +809,7 @@
 		(path "/9ca979cd-25fb-4faf-8ccd-bd26d066d55e")
 		(sheetname "/")
 		(sheetfile "kv4p-ht.kicad_sch")
-		(attr smd)
+		(attr smd dnp)
 		(fp_line
 			(start -0.14058 0.51)
 			(end 0.14058 0.51)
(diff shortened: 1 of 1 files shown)


dkaukov:
Hey @SmittyHalibut — while you're here, might be worth considering adding a low-pass filter on the DAC output:
image

SmittyHalibut:
I'm checking with JLCPCB whether I can cancel my order, or whether its too late.  If it's not too late, I'll add this.

SmittyHalibut:
image

Because of COURSE it would be in that very tight corner...  *sigh*

SmittyHalibut:
JLCPCB was able to cancel the previous order.  I uploaded this new version, with the 22nF LPF cap.  So v2.0e has the #308 filter.

VanceVagell:
Glad to hear you were able to revise your order! Thanks for all the great updates to the design.

PR #308, opened 2025-07-31
Switch from internal DAC to I2S PDM output

dkaukov:
Switched audio output from ESP32’s built-in DAC to I2S in PDM mode. This seems to give better audio quality, even without a low-pass filter. There’s still a slight click at the end of TX on the 2.0c board, but it’s subtle and probably fine.

This also prepares us for newer ESP32 chips that don’t have a DAC, and lets us map the audio output to any pins.

### DIFF
--- a/microcontroller-src/kv4p_ht_esp32_wroom_32/txAudio.h
+++ b/microcontroller-src/kv4p_ht_esp32_wroom_32/txAudio.h
@@ -25,7 +25,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #include "protocol.h"
 
 bool txStreamConfigured = false;
-AnalogAudioStream out;
+I2SStream out;
 AudioInfo txInfo(AUDIO_SAMPLE_RATE, 1, 16);
 OpusAudioDecoder txDec;
 EncodedAudioStream txOut(&out, &txDec); 
@@ -37,10 +37,11 @@ const uint16_t RUNAWAY_TX_SEC = 200;
 void initI2STx() {  
   auto config = out.defaultConfig(TX_MODE);
   config.copyFrom(txInfo);
-  config.adc_pin = hw.pins.pinAudioOut;
-  config.is_blocking_write = true;
+  config.pin_data = hw.pins.pinAudioOut;
+  config.pin_ws = 27;
   config.use_apll = true;
   config.auto_clear = false;
+  config.signal_type = PDM;
   out.begin(config);
   // configure OPUS additinal parameters
   txDec.setAudioInfo(txInfo);


VanceVagell:
Thanks Dmitry. Could you please confirm that APRS rx still works well?
That's my main concern with any changes in the audio chain. During
prototyping, it was very sensitive to audio quality.

dkaukov:
Thanks! I’ve tested RX on a limited set of 1.7 boards and it seems to be working fine — APRS decoding looks good. That said, I think we’ll need you or @SmittyHalibut  to test on the 2.xx variants to be sure everything holds up there too.

lyusupov:
> The filter shown above (10nF and 1.5k) filters to about 10kHz.

Correct 
the R=10K + C=1.5nF filter in T-TWR V2.x is focused on 10 kHz bandwidth
to meet APRS and, possibly, digital voice specs.

There are actually 2 PDM outputs in ESP32-S3 based T-TWR V2.1 _( only one in T-TWR V2.0 )_
1) one for SA868 input
2) another one to play a voice or music on the built-in speaker

both of them use the same R and C filter values for simplicity.

SmittyHalibut:
Post copied from #307 :

image

Relevant bit: IO25 is the mic audio out.  C9 blocks DC.  R2/R3 are a voltage divider to bring the level down to mic-level-ish.  C24/C34 are RF shunt caps to keep any 145MHz/440MHz RF out of the mic line coming back from the radio to the microcontroller.

I added C38, 22nF.  C38 and R3 form a LPF at 6kHz (1/2PiRC).  Well, R3 in parallel with whatever load the radio provides which is much higher than 1.2k so we can ignore it.  R3 will be dominant in draining C38 and therefore sets the frequency.  Worst case, C38 and C24 in parallel are 24.2nF, which brings the -3dB point down to 5.5kHz.  Still above the ~4kHz the radio is doing.

dkaukov:
@SmittyHalibut  @VanceVagell , I’ve tested on my existing hardware:

1. I have multiple 1.7 UHF boards and a single 2.0c VHF board.
2. I can test subjective audio quality on both VHF and UHF using a normal radio.
3. I can only test APRS RX on UHF, between the 1.7 boards.

Results so far:

* Audio quality is subjectively not worse — and possibly better — across the board.
* APRS decoding works fine on UHF between 1.7 boards.
* On the 2.0c VHF board, there’s a subtle click at the end of TX.
* The 1.7 UHF boards do not have that click.

I think we’ll still need more data to confirm that everything is fully OK. 😉

SmittyHalibut:
image

lyusupov:
As a side note -
use  of `classic` ESP32 in a new hardware design  looks ridiculous in year 2025...
Have you thought about to transit on ESP32-S3 or similar recent Espressif SoC to prove that V2 hardware design of this project is really significant step over V1 ?

SmittyHalibut:
🙄  🤣 

When Vance started this project, the classic ESP32 was the most commonly available development module with a built-in DAC.  It's what he had access to, so its what he used.  Since then, the priority has been to make things work, and make things work more reliably.  Switching out something that works for a newer thing that doesn't (yet) work, hasn't been a priority.

Issue #73, opened 2024-11-02
[Android] Display app and firmware version at bottom of Settings page


VanceVagell:
We should display both versions to aid in future debugging when people report issues.

The app version will be like: x.x.x
The firmware version will just be a number like: x (these are just ascending integers for each firmware release)

For example:

App v1.0.0
Firmware v1

VanceVagell:
Implemented in next version.

PR #311, opened 2025-08-05
Version info

dkaukov:
WIP for https://github.com/VanceVagell/kv4p-ht/issues/73


VanceVagell:
Re-reviewing PRs to see where they're at. Upon further consideration, I'd prefer to continue defining versions explicitly in the source and config files rather than indirectly via GitHub.

Please action or close out this PR when you get a chance, so we can clean up the PR list.

PR #316, opened 2025-08-12
Experimental AFSK decoder/encoder

dkaukov:
1) Should use less CPU
2) Decoding recorded packets with clicks
3) Uses CRC16 bit recovery

Current results with TNC TestCD: 1039/988

https://github.com/dkaukov/afsk-java


VanceVagell:
Testing now...

Before I forget, note that we should also remove the old AFSK library from the project, if we go with this new way. Also, I'd prefer we embed the actual decoder into our project (in its own package) to avoid dependencies that can break later outside this project's control. I know the AFSK decoder is your own project, but I'd hold the same bar for any 3rd party dependency.

I believe we may have a few truly external dependencies, but they're all for long-lived stable dependencies. I wouldn't be averse to pulling a fork of them into this project as well, but it's less risky to just leave them.

VanceVagell (changes requested):
This seems to work great, it had no issues decoding on my v1.x or v2.x test devices.

There's tons of log spam though, that looks like:
PLL phase=0.29116952, step=0.0251275, err=-0.26704663, locked=true

Could we please supress that, or make it an option that can be passed in to the library? Otherwise the kv4p HT log will be mostly that. :)

dkaukov:
Hi @VanceVagell, thanks for the feedback!

Just a couple quick notes:

* The new library currently only includes the decoder — no encoder or packet parser yet.
* Logging is already fixed — the PLL spam is disabled by default now.

As for keeping it separate, I’d prefer that for now. The main kv4p project doesn’t have much in the way of testing, and having it standalone makes it much easier for me to run large test sets (like megabytes of recordings) to tune and validate the decoder.

Totally understand the concern about stability — since it’s my own library, I’ll keep it stable and versioned. Happy to find a middle ground if needed!

VanceVagell:
Ah ok, understood that it's a WIP. Certainly a valuable project in the
technical ham world, there are so few options for working with packets in
Java.

And it's your project so of course if you think keeping it separate is the
right call, that's your call. Once you've got it fully fledged, maybe we
can pull it in as a compiled library so I can keep a fixed copy in kv4p HT
for stability. But for now I'm ok referring to it like this as you mature
it.

Thanks for improving this!

VanceVagell:
If it improves the reliability of packets, I'd be thrilled to move to it.
Most of what I really want to build in the future is packet based.

But I'd prefer to move to it wholesale and remove the prior library rather
than have one library for decode and another for encode.

VanceVagell:
Quick reply, before I get a chance to send you a recording.

I've never seen a corrupted ax.25 decode on the current library, regardless
of hardware.

dkaukov:
Hi @VanceVagell , any luck getting that recording when you have a chance? It’d help a lot to dig into what’s happening on the v1.7b. Also, I noticed the PR still shows as “changes requested,” but I can’t see an actual request left on the thread — did I miss something?

PR #328, opened 2025-08-25
Cleanup permissions request code:

dkaukov:
1) Created helper ensurePermissions(perms, callback) 
2) Moved all startup code to asunc execution chain 
3) If something not granted - show message.


VanceVagell:
Works great, thanks!

Issue #330, opened 2025-08-29
[Website] fix NVS flasher


VanceVagell:
Test if it's working.

In GitHub, the bin files are symlinks, did they upload correctly?

We had one report of a 2.x user who could not flash NVS, they got an error dialog from the web flasher.

VanceVagell:
Also see #336

PR #335, opened 2025-08-29
Follow symlinks

dkaukov:
https://help.dreamhost.com/hc/en-us/articles/216456227--htaccess-overview

> Typically, DreamHost's shared hosting environments are configured with AllowOverride All or a similar setting for user directories, allowing common .htaccess directives to function as expected. This enables users to control aspects like URL rewriting (using mod_rewrite), custom error pages, access control, and authentication within their specific website directories.

May fix: https://github.com/VanceVagell/kv4p-ht/issues/330

### DIFF
--- a/website-src/hwconfig/.htaccess
+++ b/website-src/hwconfig/.htaccess
@@ -1,0 +1,1 @@
+Options +SymLinksIfOwnerMatch


dkaukov:
@VanceVagell Nah, the flasher literally writes the filename into flash at 0x9000. When the firmware starts up it sees that as a corrupted NVS block and resets back to defaults. So it only “works” in your case because you’re not changing anything away from the defaults.

PR #337, opened 2025-08-30
Added my own case designs for version 2.x PCB

rivitir:
Added a couple of my own 3D printable case designs for the 2.x version.


VanceVagell:
Thanks!

Issue #378, opened 2026-02-14
[Android + Firmware] Changing to another app then returning to kv4p HT app may show "Need to flash firmware" message


VanceVagell:
I'm regularly seeing when I have kv4p HT running in the background and then I return to it after using other apps, it will try to reconnect to the radio (not sure why since it has been working all along in the background), but it fails and claims the firmware needs to be flashed. Restarting the app fully fixed it.

I suspect we're not gracefully connecting to the background service and it's instead being recreated or something like that. Need to debug this one so we don't get misleading errors when swapping back to the app.

PR #379, opened 2026-02-17
First step toward ESP32 AFSK modem: dual-path APRS TX/RX for field testing

dkaukov:
This is the first step (if all goes well) toward moving the AFSK modem into firmware.

  Big picture: we want kv4p to work as a TNC terminal. A lot of people have asked for that.

  ## Changes in this PR

  - Added ESP32 support for AX.25 TX/RX over the app<->firmware protocol.
  - Added AFSK modem pieces in firmware:
      - ESP32 can now modulate AX.25 for TX
      - ESP32 can demodulate AFSK on RX and send decoded packets back to Android
  - Added a new app setting: APRS TX encoder (Android or ESP32).
  - App now shows where APRS packets were decoded:
      - [A] Android
      - [E] ESP32
      - [A+E] both
  - Added duplicate handling so we don’t double-process packet side effects when both paths decode the same packet.

  ## Why

  Test modem in the wild and build confidence.


android-src/KV4PHT/app/src/main/res/values/strings.xml
```diff
@@ -98,6 +98,7 @@
     <string name="position_beacon_message_2">" position on active frequency"</string>
     <string name="settings">Settings</string>
     <string name="beacon_my_position">Beacon my position</string>
+    <string name="aprs_tx_encoder">APRS TX encoder</string>
```
VanceVagell: Any string we add in English also needs a Chinese translation in https://github.com/VanceVagell/kv4p-ht/blob/main/android-src/KV4PHT/app/src/main/res/values-zh-rCN/strings.xml . Please use an LLM to translate it, so you can provide more context so the strings make sense in Chinese (e.g. don't just translate this string but say something like "Translate this English UI string to Chinese, it will appear in an app settings screen and is used to choose which backend method is used to encode APRS ham radio text messages and location beacons.").

dkaukov:
Also, it should unblock others — once it’s merged, anyone can pick it up and keep iterating from there.

dkaukov:
Hi @VanceVagell — makes sense to focus on stability first.

One thought: I *think* this PR already fixes or at least improves the flaky handshake. I’ve been running it on my phone and I can’t reproduce the “Connecting to radio…” loop or the reconnect-on-focus issues anymore.

If you’ve got a minute, could you try this branch and see if #378 still happens for you?

Branch: [https://github.com/dkaukov/kv4p-ht/tree/esp32-afsk-integration](https://github.com/dkaukov/kv4p-ht/tree/esp32-afsk-integration)

If it does look better, I’d much rather merge this as-is than try to tease the handshake changes out into a separate PR.

dkaukov:
Hi @VanceVagell, see: https://github.com/VanceVagell/kv4p-ht/pull/382

android-src/KV4PHT/app/schemas/com.vagell.kv4pht.data.AppDatabase/7.json
```diff
@@ -0,0 +1,234 @@
+{
```
VanceVagell: Please remove the DB schema changes from the PR, and the code which persists the APRS decoder to the DB.

Let's keep the DB agnostic of what APRS decoder is being used, because the end state is that we will only have one decoder.

If we get this PR to a point where I'm OK with us having an advanced option to test this alternative decoder, it will only be for us (and other developers) to compare and contrast. To that end, it's 100% ok to print to the log what decoder what used so we can get that information if needed, but let's not persist it to the DB.

android-src/KV4PHT/app/src/main/java/com/vagell/kv4pht/data/migrations/MigrationFrom6To7.java
```diff
@@ -0,0 +1,34 @@
+/*
```
VanceVagell: Similar to previous comment, please revert DB changes.

android-src/KV4PHT/app/src/main/java/com/vagell/kv4pht/data/AppSetting.java
```diff
@@ -41,9 +41,10 @@ public class AppSetting {
     public static final String SETTING_HIGHPASS = "highpass";
     public static final String SETTING_LOWPASS = "lowpass";
     public static final String SETTING_DISABLE_ANIMATIONS = "disableAnimations";
-    public static final String SETTING_APRS_POSITION_ACCURACY = "aprsPositionAccuracy";
-    public static final String SETTING_APRS_BEACON_POSITION = "aprsBeaconPosition";
-    public static final String SETTING_CALLSIGN = "callsign";
+    public static final String SETTING_APRS_POSITION_ACCURACY = "aprsPositionAccuracy";
+    public static final String SETTING_APRS_BEACON_POSITION = "aprsBeaconPosition";
+    public static final String SETTING_APRS_TX_ENCODER = "aprsTxEncoder";
```
VanceVagell: I am still reviewing the PR, but does the PR only add the alternative APRS backend for encoding, not decoding?

The comment on the PR describing what it does suggests that both tx/rx are supported. If so, why is this setting "aprsTxEncoder" instead of "aprsEncoder"?

android-src/KV4PHT/app/src/main/java/com/vagell/kv4pht/radio/Protocol.java
```diff
@@ -62,7 +63,8 @@ public enum RcvCommand {
         COMMAND_HELLO(0x06),            // [COMMAND_HELLO()]
         COMMAND_RX_AUDIO(0x07),         // [COMMAND_RX_AUDIO(int8_t[])]
         COMMAND_VERSION(0x08),          // [COMMAND_VERSION(Version)]
-        COMMAND_WINDOW_UPDATE(0x09);    // [COMMAND_WINDOW_UPDATE()]
+        COMMAND_WINDOW_UPDATE(0x09),    // [COMMAND_WINDOW_UPDATE()]
+        COMMAND_AX25_RX_PACKET(0x0A);   // [COMMAND_AX25_RX_PACKET(uint8_t decoderId, uint8_t[])]
```
VanceVagell: Please rename to COMMAND_RX_AX25_PACKET to parallel naming convention of COMMAND_RX_AUDIO.

android-src/KV4PHT/app/src/main/java/com/vagell/kv4pht/radio/RadioAudioService.java
```diff
@@ -141,6 +144,7 @@ public class RadioAudioService extends Service implements PacketHandler {
     private static final int MS_SILENCE_AFTER_DATA_MS = 700;
     private static final String MESSAGE_NOTIFICATION_CHANNEL_ID = "aprs_message_notifications";
     private static final int MESSAGE_NOTIFICATION_TO_YOU_ID = 0;
+    private static final long PACKET_SIDE_EFFECT_DEDUP_MS = 15000L;
```
VanceVagell: Please add a descriptive comment to this one, it's hard to figure out what it is from name alone.

android-src/KV4PHT/app/src/main/java/com/vagell/kv4pht/radio/RadioAudioService.java
```diff
@@ -1215,12 +1273,30 @@ public void handlePacket(byte[] packet) {
...
             }
             // Notify callbacks about the received packet
-            callbacks.packetReceived(aprsPacket);
+            callbacks.packetReceived(aprsPacket, decoderSource, hash);
         } catch (Exception e) {
             Log.d(TAG, "Unable to parse an APRS packet, skipping.");
         }
     }
 
+    private boolean shouldEmitPacketSideEffects(@NonNull String packetHash) {
```
VanceVagell: Please add a descriptive comment to this method, it's unclear to me without deeply stepping through this logic what the purpose of this is. Also consider renaming it, "emit packet side effects" is a strange concept to me.

android-src/KV4PHT/app/src/main/java/com/vagell/kv4pht/ui/APRSAdapter.java
```diff
@@ -294,5 +298,29 @@ public void setObjName(String objName) {
             }
             textViewObjName.setText(objName);
         }
+
```
VanceVagell: As noted in previous comments, please revert the APRS messages changes (and DB changes) to display and persist the source of the decode. This is transient, we will only support one encoder/decoder in the long run after we test our options. As I mentioned in another comment, we can include in the logcat what decoder/encoder was used so that we can review that if needed.

android-src/KV4PHT/app/src/main/res/drawable/decoder_source_android.xml
VanceVagell: As commented elsewhere, doesn't this actually handle encode AND decode? Why is it referred to as "decoder" here (in this filename and other filenames here)?

android-src/KV4PHT/app/src/main/res/layout/aprs_object.xml
```diff
@@ -33,6 +33,21 @@
             android:textSize="16sp"
             android:layout_marginRight="4dp" />
 
+        <TextView
+            android:id="@+id/decoderSource"
```
VanceVagell: As noted earlier, revert.

android-src/KV4PHT/app/src/main/res/layout/aprs_position.xml
```diff
@@ -33,6 +33,21 @@
             android:textSize="16sp"
             android:layout_marginRight="4dp" />
 
+        <TextView
```
VanceVagell: Revert

android-src/KV4PHT/app/src/main/res/layout/aprs_unknown.xml
```diff
@@ -33,6 +33,21 @@
             android:textSize="16sp"
             android:layout_marginRight="4dp" />
 
+        <TextView
```
VanceVagell: Revert

microcontroller-src/kv4p_ht_esp32_wroom_32/protocol.h
```diff
@@ -55,6 +56,7 @@ enum SndCommand {
   COMMAND_RX_AUDIO       = 0x07, // [COMMAND_RX_AUDIO(int8_t[])]
   COMMAND_VERSION        = 0x08, // [COMMAND_VERSION(Version)]
   COMMAND_WINDOW_UPDATE  = 0x09,
+  COMMAND_AX25_RX_PACKET = 0x0A, // [COMMAND_AX25_RX_PACKET(uint8_t decoder, uint8_t[])]
```
VanceVagell: Similar to previous comment, rename to COMMAND_RX_AX25_PACKET to match naming convention of COMMAND_RX_AUDIO.

VanceVagell (changes requested):
This is really great work! There are a few parts I'd like to see reverted (mainly the DB persistence, and the UI indicator), but the heart of this looks solid. I'm excited to see what this will let us unlock in terms of new features and better decode reliability.

android-src/KV4PHT/app/src/main/res/values/strings.xml
dkaukov: ok

android-src/KV4PHT/app/src/main/java/com/vagell/kv4pht/data/AppSetting.java
dkaukov: For decoding, the Android and ESP32 backends run in parallel. Received frames are de-duplicated and the UI records which backend(s) produced the frame, so packets can be marked as Android, ESP32, or both.  For encoding/transmit, only one backend can be used at a time, so that part is controlled by a setting.

dkaukov:
Hi @VanceVagell — thanks! Glad the core looks solid 🙂

Re the DB bit: the packet hash is just the simplest way to dedupe when the same APRS packet comes from both decoders (Android + ESP32). It gives us a stable key to compare. And if we’re storing the hash anyway, I don’t really see the harm in also storing the source — it’s super handy for comparison.

Just to be clear how it works today: **decoders run in parallel**, and we **de-duplicate frames using the hash in the DB**. That also lets us know (and optionally display) whether a packet was decoded by **Android**, **ESP32**, or **both** — basically an A/B test in the real world.

On TX, the **encoder is switchable in settings** (software vs firmware), so only one path is used at a time.

UI indicator: totally get the concern it might confuse people. Happy to hide it behind a settings toggle so folks who don’t care never see it.

android-src/KV4PHT/app/src/main/java/com/vagell/kv4pht/data/APRSMessageDao.java
```diff
@@ -34,6 +34,9 @@ public interface APRSMessageDao {
     @Query("SELECT * FROM aprs_messages WHERE `from_callsign` = :fromCallsign AND `message_num` = :msgNum ORDER BY `id` DESC LIMIT 1")
     APRSMessage getMsgToAck(String fromCallsign, int msgNum);
 
+    @Query("SELECT * FROM aprs_messages WHERE `packet_hash` = :packetHash ORDER BY `id` DESC LIMIT 1")
+    APRSMessage getLatestByPacketHash(String packetHash);
```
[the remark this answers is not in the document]
dkaukov: It is used here: https://github.com/dkaukov/kv4p-ht/blob/esp32-afsk-integration/android-src/KV4PHT/app/src/main/java/com/vagell/kv4pht/ui/MainActivity.java#L789 to deduplicate messages.

android-src/KV4PHT/app/src/main/java/com/vagell/kv4pht/radio/RadioAudioService.java
```diff
@@ -844,6 +873,18 @@ public void setupSerialConnection() {
         }
         // Open a connection to the first available driver.
         UsbSerialDriver driver = availableDrivers.get(0);
+        if (!manager.hasPermission(driver.getDevice())) {
```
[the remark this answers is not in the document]
dkaukov: This is removed now. As it was merged to main.

android-src/KV4PHT/app/src/main/java/com/vagell/kv4pht/radio/RadioAudioService.java
dkaukov: Done. Also renamed to the `PACKET_ACTION_DEDUP_WINDOW_MS`

android-src/KV4PHT/app/src/main/java/com/vagell/kv4pht/radio/RadioAudioService.java
dkaukov: Added comment and renamed to the `shouldTriggerPacketActions()`

android-src/KV4PHT/app/src/main/res/drawable/decoder_source_android.xml
dkaukov: Decoders run in parallel; the encoder is switchable in settings.

android-src/KV4PHT/app/src/main/java/com/vagell/kv4pht/ui/SettingsActivity.java
```diff
@@ -104,6 +104,7 @@ private void populateMicGainOptions() {
 
     private void populateAprsOptions() {
         setDropdownOptions(R.id.aprsPositionAccuracyTextView, List.of("Exact", "Approx"));
+        setDropdownOptions(R.id.aprsTxEncoderTextView, List.of("Android", "ESP32"));
```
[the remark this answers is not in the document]
dkaukov: I think there’s a misunderstanding here — before I move stuff around, can we align on what this setting actually controls (and what settings we want to have overall)? 😉

microcontroller-src/kv4p_ht_esp32_wroom_32/kv4p_ht_esp32_wroom_32.ino
```diff
@@ -138,7 +138,9 @@ void doConfig(Config const &config) {
       break;
     }
   }
-  uint8_t features = (hw.features.hasHL ? FEATURE_HAS_HL : 0) | (hw.features.hasPhysPTT ? FEATURE_HAS_PHY_PTT : 0);
+  uint8_t features = (hw.features.hasHL ? FEATURE_HAS_HL : 0)
+    | (hw.features.hasPhysPTT ? FEATURE_HAS_PHY_PTT : 0)
+    | FEATURE_HAS_ESP32_AFSK;
```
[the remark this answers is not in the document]
dkaukov: AFSK

android-src/KV4PHT/app/schemas/com.vagell.kv4pht.data.AppDatabase/7.json
dkaukov: See: https://github.com/VanceVagell/kv4p-ht/pull/379#issuecomment-3981324561

android-src/KV4PHT/app/src/main/java/com/vagell/kv4pht/data/AppDatabase.java
```diff
@@ -30,10 +30,10 @@ kv4p HT (see http://kv4p.com)
 /**
  * Singleton Room database for kv4p HT application.
  */
-@Database(
-    version = 6,
-    entities = {AppSetting.class, ChannelMemory.class, APRSMessage.class}
-)
+@Database(
```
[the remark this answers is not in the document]
dkaukov: See: https://github.com/VanceVagell/kv4p-ht/pull/379#issuecomment-3981324561

android-src/KV4PHT/app/src/main/java/com/vagell/kv4pht/data/migrations/MigrationFrom6To7.java
dkaukov: See: https://github.com/VanceVagell/kv4p-ht/pull/379#issuecomment-3981324561

android-src/KV4PHT/app/src/main/java/com/vagell/kv4pht/radio/Protocol.java
dkaukov: done

android-src/KV4PHT/app/src/main/java/com/vagell/kv4pht/ui/APRSAdapter.java
dkaukov: See: https://github.com/VanceVagell/kv4p-ht/pull/379#issuecomment-3981324561

android-src/KV4PHT/app/src/main/res/layout/activity_settings.xml
```diff
@@ -296,6 +296,45 @@
                     </com.google.android.material.textfield.TextInputLayout>
                 </LinearLayout>
 
+                <LinearLayout
```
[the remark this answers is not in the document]
dkaukov: Not sure I follow.

android-src/KV4PHT/app/src/main/res/layout/aprs_message.xml
```diff
@@ -50,6 +50,21 @@
             android:textSize="16sp"
             android:layout_marginRight="4dp" />
 
+        <TextView
+            android:id="@+id/decoderSource"
```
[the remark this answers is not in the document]
dkaukov: See: https://github.com/VanceVagell/kv4p-ht/pull/379#issuecomment-3981324561

android-src/KV4PHT/app/src/main/res/layout/aprs_position.xml
dkaukov: See: https://github.com/VanceVagell/kv4p-ht/pull/379#issuecomment-3981324561

android-src/KV4PHT/app/src/main/res/layout/aprs_unknown.xml
dkaukov: See: https://github.com/VanceVagell/kv4p-ht/pull/379#issuecomment-3981324561

android-src/KV4PHT/app/src/main/res/layout/aprs_weather.xml
```diff
@@ -33,6 +33,21 @@
             android:textSize="16sp"
             android:layout_marginRight="4dp" />
 
+        <TextView
```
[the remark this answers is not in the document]
dkaukov: see: https://github.com/VanceVagell/kv4p-ht/pull/379#issuecomment-3981324561

microcontroller-src/kv4p_ht_esp32_wroom_32/protocol.h
dkaukov: done

android-src/KV4PHT/app/src/main/res/layout/aprs_object.xml
dkaukov: See: https://github.com/VanceVagell/kv4p-ht/pull/379#issuecomment-3981324561

microcontroller-src/kv4p_ht_esp32_wroom_32/kv4p_ht_esp32_wroom_32.ino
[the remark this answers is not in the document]
dkaukov: My rough next steps would be:

* Clean up the Java side (encoder/decoder + the A/B plumbing) so it’s not fighting the firmware path.
* Add a KISS mode over USB serial (SLIP/KISS framing). I’m leaning toward replacing the current DEADBEEF protocol with KISS + a couple custom extensions.
* Add settings + persistence (NVS on ESP32 + app UI) for the “TNC terminal” basics: frequency, power, squelch/bandwidth, etc.
* Find a lighter replacement for OPUS so we can enable the radio reliably (and later bring BT back too).

microcontroller-src/kv4p_ht_esp32_wroom_32/kv4p_ht_esp32_wroom_32.ino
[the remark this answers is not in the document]
VanceVagell: Thanks. To make sure I understand, is it the case that FEATURE_HAS_ESP32_AFSK is intended to mean that the firmware is expected to be capable of other AFSK-based encodings in the future, such as IL2P? If so then I think this name makes sense.

That said, since this is defined in firmware and is not dependent on the physical hardware, I'm not sure why we need this as a feature flag at all. We expect the firmware to always have all features so long as they're not dependent on physical hardware (e.g. a button).

android-src/KV4PHT/app/src/main/java/com/vagell/kv4pht/ui/SettingsActivity.java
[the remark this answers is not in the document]
VanceVagell: Based on our discussion so far, the changes in this PR seem to be primarily to introduce a new AX.25 encoder/decoder in the firmware. Here's my understanding from the user's perspective:

- Only one encoder can be used, so the user must choose which one they'd like to use.
- The decoder has more flexibility, since the user can select either software/hardware decoding, or both (where we de-dupe any packets that both were able to decode).

I believe you're recommending that most users be in the "both" decoder state to get a better sense of how these decoders are working in the wild. If this is accurate, then what I'm saying in this comment is that the settings screen needs adjustments from what's in this PR, for a few reasons:

- This is an AX.25 setting, not an APRS setting
- It's an advanced setting the vast majority of users do not need to understand, and also one we expect to remove in the future

As a result of this, please include it in the UI as such, at the end of the Advanced section (which I've excerpted here for clarity):

Advanced
Bandwidth [25kHz]
Min 2m TX frequency [144MHz]
Max 2m TX frequency [148MHz]
Min 70cm TX frequency [420MHz]
Max 70cm TX frequency [450MHz]
Mic gain boost [None]
AX.25 encoder [Software] (options: Software, Firmware)
AX.25 decoder [Both] (options: Both, Software, Firmware)

The options shown (Software, Both) are the default values if the user does not change them.

(Note: Pending resolution of this comment https://github.com/VanceVagell/kv4p-ht/pull/379/changes#r2925009350 we may need to default decoder to "Software" if there's a risk of duplicate decodes with slightly different errors, so that people don't experience a regression in the experience of APRS chat.)

Note that if the user selects "Firmware" for decoder, you need not disable decoding in the firmware, just ignore any packets it reports.

android-src/KV4PHT/app/src/main/res/layout/activity_settings.xml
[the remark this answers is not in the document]
VanceVagell: See: https://github.com/VanceVagell/kv4p-ht/pull/379/changes#r2924983774

android-src/KV4PHT/app/src/main/java/com/vagell/kv4pht/data/APRSMessageDao.java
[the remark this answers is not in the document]
VanceVagell: Thanks, that's helpful. I'm trying to figure out how the hash is generated, I'm not seeing it in the PR, but am probably just missing it.

That said, what I'm really concerned about is what happens if the decoders decode the same message slightly differently. I believe our software decoder will generally fail rather than partially decode a noisy signal, whereas the new firmware decoder is more lenient with passing along partial decodes. As a result, they may generate different hashes and appear as duplicate (but subtly different) messages in the user's UI.

Could you please confirm what would actually happen in this situation? In my limited testing of the new firmware decoder (the last time you sent it over, not as of this PR) I was seeing some damaged packets decoded slightly incorrectly, I'd expect those to show up as very-similar copies after this PR.

android-src/KV4PHT/app/schemas/com.vagell.kv4pht.data.AppDatabase/7.json
VanceVagell: Some of the code review comments I just added will help me understand whether the DB changes are actually necessary or not. Reopening this thread as that decision still needs to be made.

VanceVagell (commented):
Thanks for the extra explanation, Dmitry, very helpful.

One thing I'm still not 100% clear on is whether this PR relates to "APRS" at its heart, or "AX.25". From our discussion so far and what I see in the code, I believe this is fundamentally an expansion of how we handle AX.25, and then it is _applied_ to APRS messaging. This is primarily important in terms of settings names and how we talk about it in the UI (at least until we remove those settings in the future and use a single encoder/decoder that works best). Could you please clarify that?

As an aside, to make larger changes/reviews like this go more smoothly in the future, it would be helpful if we discussed the overall goals and architecture before you send over a PR, just to get on the rough same page. Then we can just focus on the implementation instead of "why" -- in my experience, it's not enjoyable as a software engineer to get lots of fundamental high level questions on a PR that have sweeping impact on the code you've written. A short Google Doc or even just a detailed discussion in a GitHub Issue would be fine, or whatever is easiest for you.

android-src/KV4PHT/app/src/main/java/com/vagell/kv4pht/data/APRSMessageDao.java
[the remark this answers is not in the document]
dkaukov: **Hash generation** is in `RadioAudioService.java` — GitHub collapses that file by default because it’s big, so you have to hit “Expand”. It’s here:
[https://github.com/dkaukov/kv4p-ht/blob/33386c06345ce3ac77dc3ea1e3e1c4f767b138b0/android-src/KV4PHT/app/src/main/java/com/vagell/kv4pht/radio/RadioAudioService.java#L1447](https://github.com/dkaukov/kv4p-ht/blob/33386c06345ce3ac77dc3ea1e3e1c4f767b138b0/android-src/KV4PHT/app/src/main/java/com/vagell/kv4pht/radio/RadioAudioService.java#L1447)


On the “same packet decoded slightly differently” concern: I don’t think that can happen here, because **both decoders only emit a frame if the AX.25 CRC passes**. If the bytes differ, CRC would fail and the frame should be dropped. So in practice you either get:

* exact same bytes from both → same hash → deduped, marked S+F, or
* one decoder fails CRC → you just get the one good decode.

I think you might be remembering the older experiments / **afsk-java bit-flip recovery** cases, where we intentionally try to “repair” frames and you can get near-miss variants. The ESP32 decoder in this PR **doesn’t do bit-flip recovery**, so it shouldn’t be passing along “partial decodes”.

dkaukov:
Hi @VanceVagell,

Not sure it’s super convenient to discuss the “big picture” stuff on a specific line thread, so I’m duplicating this as a general comment.

Yep — this is AX.25/AFSK at its core (APRS chat just uses it today). I’ve updated the settings to reflect that (AX.25 encoder/decoder in Advanced).

Hash generation is in `RadioAudioService.java` — GitHub collapses that file by default because it’s large, so you need to hit **Expand**. It’s here:
[https://github.com/dkaukov/kv4p-ht/blob/33386c06345ce3ac77dc3ea1e3e1c4f767b138b0/android-src/KV4PHT/app/src/main/java/com/vagell/kv4pht/radio/RadioAudioService.java#L1447](https://github.com/dkaukov/kv4p-ht/blob/33386c06345ce3ac77dc3ea1e3e1c4f767b138b0/android-src/KV4PHT/app/src/main/java/com/vagell/kv4pht/radio/RadioAudioService.java#L1447)

On the “same packet decoded slightly differently” concern: I don’t think that should happen here, because **both decoders only emit a frame if the AX.25 CRC passes**. If the bytes differ, CRC should fail and the frame gets dropped. So in practice it’s either:

* same bytes from both → same hash → deduped, marked **S+F**, or
* one side fails CRC → you just get the one good decode.

I think you might be remembering the older **afsk-java bit-flip recovery** experiments, where we intentionally try to “repair” frames and you can get near-miss variants. The ESP32 decoder in this PR **doesn’t** do bit-flip recovery, so it shouldn’t be passing “partial decodes”.

Also, I addressed the settings request — we now have:

* **AX.25 decoder** [Both] (Both / Software / Firmware)
* **AX.25 encoder** [Software] (Software / Firmware)
  
…at the end of **Advanced**.

android-src/KV4PHT/app/src/main/java/com/vagell/kv4pht/ui/SettingsActivity.java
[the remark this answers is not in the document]
dkaukov: image

dkaukov:
Hi @VanceVagell, just checking back on this when you have a chance. I think the latest feedback is addressed now, so I’d appreciate another look.

microcontroller-src/kv4p_ht_esp32_wroom_32/kv4p_ht_esp32_wroom_32.ino
VanceVagell: Any change to the firmware should increment the firmware version, please fix this.

It should also update the Android app to bundle the new firmware binary so people can flash it, but that can be a follow up PR, since there could be more firmware changes before we prepare the next release.

android-src/KV4PHT/app/src/main/java/com/vagell/kv4pht/data/migrations/MigrationFrom6To7.java
VanceVagell, line 1: (Comment is unrelated to this file, I just need to put it somewhere since it relates to another file not in the PR yet. There's probably a better way to do this...)

Please update the versionCode to 52 and the versionName to "1.9.9.4" in build.gradle in this PR. The PR currently has a version that's behind HEAD, so you'd need to at least fix that and we may as well just get the new version info in there now while you're editing it.

android-src/KV4PHT/app/src/main/res/layout/aprs_position.xml
VanceVagell: I've been testing out this PR on my device, and we definitely need to hide these S/F indicators on the messages by default. 

I agree it will be helpful for me and you (and any other developers who want to) to turn these on to experiment with the various APRS settings to test the behavior, but they won't add value for 95% of people who use the app, and they'll waste mental energy trying to figure out what it means since it's non-standard with any other radio.

A couple options:
* Add a simple static variable in the source (like SHOW_AX25_DEBUG_UI) or something, that defaults to false, and only show the indicators if this was set to true at compile time.
OR
* Add a new setting in Advanced called "Developer mode", which defaults to Off, and if it's On, then this UI (and any other future helpful debug UIs we want to add) can be displayed. I know this adds a bunch to this PR, so totally fine if you prefer the latter.

VanceVagell (changes requested):
I tested out the PR on-device and with another APRS radio. Everything seems to work well. Just a couple final comments left, then we can merge this.

Sorry for the slow reviews, I wish we had more Android developers in the project to help out! I appreciate your contributions.

microcontroller-src/kv4p_ht_esp32_wroom_32/kv4p_ht_esp32_wroom_32.ino
dkaukov: done

android-src/KV4PHT/app/src/main/java/com/vagell/kv4pht/data/migrations/MigrationFrom6To7.java
dkaukov: done

android-src/KV4PHT/app/src/main/res/layout/aprs_position.xml
dkaukov: I get the UX concern, but adding a hide flag / developer-mode gate just adds another layer of complexity to a temporary experiment without really solving the core issue.

The whole point of this PR is field A/B. If the source indicator is hidden for most users, we lose most of the feedback value while still carrying all the dual-path complexity. If we don’t want users to see which path decoded a packet, then the cleaner option is to skip the dual-path A/B and just switch to firmware.

VanceVagell:
Hey Dmitry. I'm leaning towards us just moving to the new APRS backend you're working on wholesale, since it will unlock KISS TNC support which will be really great. I'm just testing it in the field for a bit to make sure it seems comparable to what we have now.

If so, we can remove the UI indicators and strip out the old APRS stuff as well (could be in a separate PR).

Will update here within the week with how the testing is going.

PR #382, opened 2026-02-27
Switched Android USB connection management from event-driven reconnect triggers to a periodic polling loop.

### LINKED ISSUE #378: [Android + Firmware] Changing to another app then returning to kv4p HT app may show "Need to flash firmware" message

dkaukov:
### Why

  The previous event-based flow (start, USB attach/detach, run-error callbacks) could overlap and race, causing repeated connect attempts, flaky handshakes, and noisy UI behavior.

  ### What changed

  - Replaced event-triggered reconnect orchestration with periodic connection polling.
  - Added ConnectionController to centralize retry timing and attempt gating.
  - RadioAudioService now delegates connect scheduling to the controller and focuses on transport/handshake operations.

  May fix: https://github.com/VanceVagell/kv4p-ht/issues/378 and https://github.com/VanceVagell/kv4p-ht/pull/379#issuecomment-3963407571

### DIFF
--- a/android-src/KV4PHT/app/src/main/java/com/vagell/kv4pht/radio/ConnectionController.java
+++ b/android-src/KV4PHT/app/src/main/java/com/vagell/kv4pht/radio/ConnectionController.java
@@ -21,10 +21,11 @@ final class ConnectionController {
             if (!isConnectionReady.getAsBoolean() && !attemptActive) {
                 attemptActive = true;
                 attemptConnect.run();
+                if (!running) {
+                    return;
+                }
             }
-            if (running) {
-                handler.postDelayed(this, periodMs);
-            }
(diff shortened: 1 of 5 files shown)


VanceVagell:
Thanks for sending this over, @dkaukov ! It does indeed seem to fix #378 from my limited testing this afternoon, I will continue to test it and report back if any further issues seen. Planning to include this in next release.

Issue #388, opened 2026-03-19
RX audio has periodic 5Hz clicking artifact caused by I2S 2% oversample


ukbodypilot:
## Bug Description

RX audio from the KV4P HT contains a periodic clicking/popping artifact at approximately **5Hz (every ~200ms)**. The artifact is present in the raw Opus-decoded output regardless of the receiving application — it originates in the ESP32 firmware's RX audio encoding pipeline.

## Root Cause

In `rxAudio.h`, the I2S ADC is configured with a 2% sample rate oversample:

```cpp
config.sample_rate = AUDIO_SAMPLE_RATE * 1.02; // 2% over sample rate to avoid buffer underruns
```

However, the Opus encoder is configured at the nominal rate (`AUDIO_SAMPLE_RATE = 48000`):

```cpp
AudioInfo rxInfo(AUDIO_SAMPLE_RATE, 1, 16);  // 48000 Hz
```

This creates a rate mismatch:
- **I2S ADC reads at**: 48,960 Hz (48000 × 1.02)
- **Opus encoder consumes at**: 48,000 Hz
- **Excess per 200ms**: 48,960 × 0.2 − 48,000 × 0.2 = **192 extra samples**
- **Period**: 200ms = exactly **5 Opus frames** (40ms each)

Every 5 frames (~200ms), the accumulated 192 extra samples cause the AudioTools internal buffer to wrap around or overflow, creating a sample discontinuity that the Opus encoder captures as a click.

## Evidence

Captured raw Opus frames directly from the ESP32 serial output (bypassing any host-side processing) and decoded to WAV. Analysis shows:

- **Sample discontinuities >1000**: 2,501 per 2 seconds of audio
- **DC offset jumps at Opus frame boundaries**: up to 288 (stdev=100)
- **5Hz energy in envelope spectrum**: 153,787 (dominant artifact frequency)
- **Measured frame rate**: 24.75 fps (confirming the I2S feed rate drives the Opus frame pacing)

The artifact is clearly audible as a regular clicking pattern during voice reception.

## Proposed Fix

Remove the 1.02 multiplier — the `use_apll = true` setting already provides a precise clock source:

```cpp
// Before:
config.sample_rate = AUDIO_SAMPLE_RATE * 1.02; // 2% over sample rate to avoid buffer underruns

// After:
config.sample_rate = AUDIO_SAMPLE_RATE; // Use exact rate — APLL provides precise clock
```

[...]

PR #389, opened 2026-03-19
Fix: remove I2S 2% oversample causing 5Hz RX audio click

ukbodypilot:
## Summary

- Remove the `* 1.02` I2S sample rate multiplier in `rxAudio.h` that causes a periodic 5Hz clicking artifact in RX audio
- The APLL (`use_apll = true`) already provides a precise clock, making the oversample unnecessary

## Problem

The I2S ADC runs at 48,960 Hz while the Opus encoder consumes at 48,000 Hz. Every 200ms (~5 Opus frames), 192 extra samples accumulate and cause a buffer discontinuity that Opus encodes as an audible click.

## Evidence

Raw Opus capture analysis (bypassing all host processing):
- 2,501 sample discontinuities >1000 per 2 seconds
- 5Hz dominant artifact in envelope spectrum
- DC offset jumps up to 288 at Opus frame boundaries

See issue #388 for full analysis methodology and data.

## Test plan

- [ ] Build and flash patched firmware
- [ ] Compare raw Opus decode before/after for discontinuity count
- [ ] Verify no I2S buffer underruns (the original reason for the oversample)
- [ ] Confirm voice audio quality improvement by listening test

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/microcontroller-src/kv4p_ht_esp32_wroom_32/rxAudio.h
+++ b/microcontroller-src/kv4p_ht_esp32_wroom_32/rxAudio.h
@@ -95,7 +95,7 @@ void initI2SRx() {
   config.use_apll = true;
   config.auto_clear = false;
   config.adc_pin = hw.pins.pinAudioIn;
-  config.sample_rate = AUDIO_SAMPLE_RATE * 1.02; // 2% over sample rate to avoid buffer underruns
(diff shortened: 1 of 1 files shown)


ukbodypilot:
Closing — the simple fix made the 5Hz artifact worse. See issue #388 for updated analysis. The root cause needs deeper investigation into the I2S DMA ↔ Opus encoder buffering interaction.

PR #392, opened 2026-04-21
Use firmware AX.25 backend unconditionally

dkaukov:
### Description:

Moves APRS AX.25 encode/decode to the ESP32 firmware path by default.

### Includes:

- ESP32 AX.25 RX/TX protocol commands
- Firmware AFSK decode/tap and AX.25 TX modulation
- Android RadioAudioService routing for firmware-decoded AX.25 packets
- Android APRS TX via firmware AX.25 command
- Firmware version/package update to v16

Alternative to the: https://github.com/VanceVagell/kv4p-ht/pull/379


VanceVagell:
Awesome, thank you @dkaukov ! I've been testing the firmware version for a bit and all seems well so far. I'm hoping to make some progress on software APRS features as well, and we'll bundle all this stuff in a big 2.0 Android app release.

PR #393, opened 2026-04-28
Improve Bluetooth TX audio handling and add RX gain boost

pioneer:
Improves Android Bluetooth headset support by moving mic capture into the radio service, adding better audio-focus handling for TX/RX, and introducing a configurable RX gain boost setting for low-volume headsets.

Tested with a Bluetooth headset: TX works in background, music resumes after radio audio stops, PTT interrupts background audio correctly, and RX volume can now be boosted from Settings.

**WARNING: This is a completely vibe-coded PR!** Although I'm an experienced developer (Python), I'm not that proficient in making apps for Android, reviewed the code to my understanding, as well as tested it on my phone. **Please have a close look!**

_My motivation was to be able to use the kv4p radio on ride, while my phone is anyway connected to the helmet headset. For this, I needed the app to work in background and behave correctly when music is being played._


VanceVagell (changes requested):
Thanks for this, I think people who use BT headsets will find this very helpful! I also look forward to using it in my car, since right now the radio audio competes with my music. :)

However, this PR contains two separate changes:
* Better BT support
* Audio gain boost

Please remove the audio gain boost part from this PR. 

Although it may be helpful in your situation, the audio volume is already carefully optimized for various kv4p HT radios to ensure it is accurate, and especially so that data transmissions (like AX.25 and APRS) work properly. The low audio you're experiencing is almost certainly either your Android device and/or your BT device not letting you set the volume as high as you'd like. We have not had other reports of this issue, and I haven't heard it on any of my test devices.

Also, please fix the issues reported by sonarqubecloud on the PR.

Finally, can you please confirm you've tested this:
* Tx/rx with a headset (using another radio to listen)
* Tx/rx without a headset (using another radio to listen)

pioneer:
Done:
- SonarQube issue
- moved TX/RX gain feature into another branch
- merge with main

Gonna build an APK and test in the mentioned scenarios, and will let you know.

pioneer:
Still not sure what's better to do with the music being played when a TX/RX appears, either pause (AUDIOFOCUS_GAIN_TRANSIENT, current behaviour) or lower the volume (AUDIOFOCUS_GAIN_TRANSIENT_MAY_DUCK), will test at least how it feels to me. Same for hardcoded 1.5s wait for no RX, maybe this number can be adjusted a bit. Although, my initial experience was ok for both, so maybe no urgent change is needed.

pioneer:
Now I can confirm that both:
- TX/RX with a headset (using another radio to listen)
- TX/RX without a headset (using another radio to listen)

work well with my phone and the helmet headset.

Observations:
- the app can be in the background and TX/RX still works
- with a headset, there's a slight but noticeable delay between pressing PTT and the actual audio I can hear over the radio (likely due to BT handshake and switching BT profiles)
- the above delay is not present without the headset
- music playback stops on either RX (audible) or TX, and resumes 1.5s after PTT release or after 1.5s of no audible RX
- music, when resumed, plays for a bit with low-quality audio because Bluetooth profiles are being switched from the phone profile back to the music one (from SCO/HFP back to A2DP)
- the RX volume is OK when no music is played; with music, the music volume exceeds the RX volume on my phone, and I couldn't find a way to adjust those volumes separately, so either the music is too loud, or the RX is too quiet -- that's why I keep thinking the RX gain boost is still usable (I believe there were no reports of such a bug just because nobody used the radio along with music playing)

pioneer:
Improved a bit the BT latency mentioned above ("with a headset, there's a slight but noticeable delay...") under "Low-latency Bluetooth mic" setting, added setting whether to pause music on RX/TX, or make it quieter in background.

VanceVagell:
Awesome, thank you for the further updates! There was a pretty big architectural change in another PR, could you please merge and resolve any conflicts? Hopefully it doesn't affect too much of your PR.

pioneer:
Yes, I'm following, exciting changes, will do my side soon.

VanceVagell:
v2.0.0.0 is going out tonight to beta users, so this feature won't make it into that version, but i look forward to including it in a future release once it's ready! thanks again

pioneer:
I'm unable to check all 2.0.0.0 changes work here, but the primary feature at least works on my phone/helmet headset. Please have a look.

Also, I still think RX gain boost setting is worth having, maybe in a separate PR.

dkaukov:
Hi @pioneer, thanks for pushing this forward. The feature direction is very useful, especially for headset/car/helmet setups, but I’d recommend splitting this into smaller PRs rather than landing the whole change at once.

Suggested split:

1. **Extract radio audio capture/playback into a dedicated audio component**
   - Move `AudioRecord`, `AudioTrack`, Opus encode/decode, TX mic frame processing, RX playback, and TX level reporting into a radio-layer audio component, e.g. `RadioAudioController`.
   - Keep `RadioAudioService` responsible for radio lifecycle, USB/protocol state, and RX/TX mode transitions.
   - Keep `MainActivity` responsible only for PTT UI state and visualization.
   - Expose clear audio methods such as `startVoiceCapture()`, `stopVoiceCapture()`, `playRxAudio(...)`, `stopPlayback()`, and `release()`.
   - Avoid adding Bluetooth routing or new audio focus policy in this PR; keep it as architecture cleanup first.

2. **Add proper audio focus management**
   - Replace per-frame focus requests during RX with an explicit focus lifecycle.
   - Request transient focus when audible RX starts or voice TX starts.
   - Release focus after a short RX/TX idle timeout so music resumes cleanly.
   - Handle focus loss by stopping RX playback and ending PTT if currently transmitting.
   - Decide separately whether the default should pause music or duck music; if ducking is configurable, that can be part of this PR or a follow-up.

3. **Add Bluetooth headset mic/routing support**
   - Add the required manifest/service permissions, including microphone foreground service type and Android 12+ Bluetooth permission handling.
   - Prefer `AudioManager.setCommunicationDevice(...)` on Android 12+.
   - Keep a legacy SCO fallback for older supported Android versions. (maybe)
   - Ensure Bluetooth routing is released on PTT end, focus loss, radio disconnect, and service shutdown.

4. **Optional follow-up: low-latency Bluetooth setting**
   - If needed, add a setting to keep/prewarm the Bluetooth communication route while the radio is connected.
   - Default should probably be off, because keeping a headset in communication/SCO mode can affect music quality and battery life.
   - This should be separate from basic headset support so the default behavior remains conservative.

I’d also suggest keeping RX gain boost out of these PRs. It changes radio/audio calibration behavior and is easier to evaluate independently from Bluetooth/focus improvements.

Overall recommendation: first land the audio architecture cleanup, then focus handling, then Bluetooth routing. That sequence gives us smaller review surfaces and makes it much easier to test regressions after each step.

VanceVagell:
Closing stale PR, thanks for sharing your custom implementation! Although we're not using it as-is, it did help with some bluetooth audio issues we wanted to improve in a different PR.

PR #408, opened 2026-05-16
Fix empty NVS HELLO seeding radio defaults

### LINKED ISSUE #404: [Android] Bandwidth setting should default to 25kHz
I suspect this is a regression as of PR #398 because we used to default to 25kHz. But after installing the new app version (after a complete app uninstall) this setting defaulted to 12.5kHz incorrectly.

Perhaps it's defaulting to that because the kv4p HT radio didn't report the bandwidth?

@dkaukov could you please take a look? I strongly suspect this is from that PR.

dkaukov:
When firmware reports a DeviceState without valid radio config, avoid copying zero-filled HELLO fields into Android desired state. Reset radio config fields to known defaults instead, preventing stale or empty NVS state from forcing 12.5 kHz bandwidth or carrying values from a previously connected module.

Fixes: https://github.com/VanceVagell/kv4p-ht/issues/404


VanceVagell:
I think this PR is a subset (and duplicate) of #409 which has been merged, correct? If so, please close the PR, otherwise please clarify what's new here.

dkaukov:
not relevan any more

PR #415, opened 2026-05-20
Gate device state reports behind host flag

### LINKED ISSUE #405: [Firmware] Radio does not fully initialize when connected to non-Android host
I'm not sure if you've implemented this yet @dkaukov so maybe this is a feature request, and not a bug.

I installed the latest firmware as of PR #398 , got it working in the kv4p HT Android app, and then tuned to 144.39 (the US APRS frequency).

I then unplugged the kv4p HT radio, and plugged it into my Windows machine. It shows up as a single COM port (perhaps intentional?) like it always did. I tried to use it with Pinpoint APRS, a Windows-based APRS client. It claims to connect to it as a KISS TNC, but I never receive messages.

**Notably, the kv4p HT radio's LED remains red throughout rather than switching to green** as it does when connected to the Android host and properly initialized by the kv4p HT app.

dkaukov:
Add an explicit desired-state flag for status reports so Android opts in when audio is open, while generic KISS TNC hosts stay quiet by default.

Related to: https://github.com/VanceVagell/kv4p-ht/issues/405


VanceVagell:
Much more robust, thank you!

PR #419, opened 2026-05-24
Filter APRS chat by destination

dkaukov:
According to spec:

https://aprspedia.com/doku.php?id=aprs_protocols%3Ainformation_field%3Amessages_bulletins_accouncements_field

### DIFF
--- a/android-src/KV4PHT/app/src/main/java/com/vagell/kv4pht/radio/RadioAudioService.java
+++ b/android-src/KV4PHT/app/src/main/java/com/vagell/kv4pht/radio/RadioAudioService.java
@@ -1669,14 +1669,14 @@ public class RadioAudioService extends Service {
     /**
      * Sends a chat message to the specified recipient.
      *
-     * @param to   The callsign of the recipient, or null for CQ.
+     * @param to   The callsign of the recipient, or null for BLN1CQ.
      * @param text The message text to send.
      * @return The message number if sent successfully, -1 on error.
      */
     public int sendChatMessage(String to, String text) {
         // Sanitize message text
         final String outText = text.replace('|', ' ').replace('~', ' ').replace('{', ' ');
-        final String targetCallsign = (to == null || to.trim().isEmpty()) ? "CQ" : to;
+        final String targetCallsign = (to == null || to.trim().isEmpty()) ? "BLN1CQ" : to;
         if (callsign.trim().isEmpty()) {
             Log.d(TAG, "Error: Tried to send message with no sender callsign.");
             return -1;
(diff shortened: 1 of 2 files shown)


VanceVagell:
Hey Dmitry. I assume this is the part of the linked spec you're referring to:

- "An APRS receiving station can specify special Message Groups, containing lists of callsigns that the station will read messages from (in addition to messages addressed to itself). Such Message Groups are defined internally by the user at the receiving station, and are used to filter received message traffic."

The key parts seem to be "An APRS receiving station _can_ specify" and "defined..._by the user_". 

A big part of APRS is monitoring local traffic and just seeing what everyone is using it for. What the spec seems to be saying is that message groups (i.e. special recipients) are a way that a client could optionally offer filtering for users who only want to see certain kinds of traffic.

I'm open to a future setting where users choose a filter (if we learn they want/need that) but I wouldn't want to decrease the amount of APRS traffic they see by default, since in many non-urban areas it will be very low already and we want them to see their device working and enjoy seeing what APRS is being used for.

But let me know if I misunderstand the intent with PR or I'm looking at the wrong part of the spec.

Also, I'm not sure we need to change "CQ" to "BLN1CQ" since the linked spec explicitly allows for "CQ" which is simpler:

- "The receiving station will read all messages with the Addressee field set to ALL, QST or CQ."

dkaukov:
Yeah, I think you’re right.

I’m also not sure hard-filtering on receive is the right place. Maybe better to receive/store everything and add an optional display filter later.

The `CQ` → `BLN1CQ` change was mostly because PinPoint APRS and some other clients seem to ignore plain `CQ`, but show `BLN1*` as broadcast/bulletin traffic.

I didn’t want to add UI changes in this PR, so I avoided custom groups and more complete bulletin handling for now.

VanceVagell:
OK I was able to reproduce that PinPoint APRS is ignoring messages to "CQ". That's a bug on their side, but as I noted previously it's more important that we work well with whatever people are using so let's make the change.

Could you please revise this PR to just have the BLN1CQ change? Not the filtering stuff.

dkaukov:
Sounds good, I’ll revise this PR to only keep the `BLN1CQ` change.

Also, it does not seem to be only PinPoint. `aprs-tnc-web` appears to have similar logic: it only treats direct messages and `BLN*` messages as chat, so plain `CQ` gets ignored there too:
https://github.com/SQ2CPA/aprs-tnc-web/blob/master/scripts/connector.ts

I still think filtering messages would be useful, but I agree it should not be part of this PR. What do you think about doing it later as a post/display filter instead of dropping messages on receive? So we would still receive/store everything, but let the chat view show “all” vs “relevant to me” later.

VanceVagell:
Yes, if we add filtering in the future we'd still store all the messages in the DB and just apply the filtering on the user-facing side. People might want different filters over time, and that will give more flexibility without locking them in to filters they had on when messages were received.

That said, APRS filtering isn't something I'd want to add now, since I expect most people have too few APRS messages, not too many.

But I'm very open to revisiting it in the future if we hear from people that they have appetite/need for even richer APRS features like this, it's not hard to add I just want to keep the UI to the essentials as much as possible.

PR #422, opened 2026-06-05
Simple SQ indicator

dkaukov:


VanceVagell:
Thanks for putting this together. To avoid churn, please check with me about UI updates, unless you're OK tweaking them afterwards with my feedback.

We can't use color for this because colorblind people can't differentiate it (particularly red/green) and it's not consistent with the approach I'm taking with the VFO area in general.

However, I do think this is a good idea! Instead of using the S-meter for this, please show a small "SQL" indicator in the lower-left of the VFO area, directly opposite where "VHF" or "UHF" is on the right side. It should show when SQL is applied (i.e. it's silent), and hide when squelch is not applied (i.e. the user hears the active frequency).

If you'd like a UX mock-up of this let me know, but I think it's pretty straightforward.

VanceVagell:
Reminder that it will also need a Chinese translation (doesn't need to be an abbreviation because the Chinese characters will already be narrow).

dkaukov:
Hi @VanceVagell, yep, makes sense.

My thinking was: this matches the current TX indicator, where I make the S-meter/band area red during TX. It is also similar to many HTs, where the backlight changes colour for RX/TX state.

But UX is your call. If you want to send a quick mock, I can adjust it. Or we can merge this and tweak it later to your liking — whatever is easier.

VanceVagell:
Thanks, here you go:

image

Left-most mock
- Show no label when squelch is off in settings, and we're in rx mode.
- Show no label when squelch is on, but currently open because of active carrier.

Middle mock
- Show "SQL" when squelch is on in settings, we're in rx mode, and currently closed (no active carrier).

Right-most mock
- Show "TX" during active transmit, along with the existing red S-meter bars (which should be full on TX). Note that "TX" is not in red so it has sufficient contrast to be easy to read.

VanceVagell:
One other note, in testing I see that the "VHF" label currently turns read during transmit, this should stay the white-ish color. Only the S-meter should change color (and the new "TX" label appears). The red color has insufficient contrast on the dark background for low-vision users.

dkaukov:
@VanceVagell done.

Maybe a TX/RX indicator would be a cleaner and more consistent way to handle this, but up to you.

VanceVagell:
The UI layout looks good, however there's one behavior issue.

Here's the repro:
1. Turn off squelch in settings
2. Press PTT
3. Notice that "TX" indicator appears (correct)
4. Release PTT
5. "TX" indicator disappears (correct)
6. "SQL" indicator briefly appears then disappears

Since squelch is off, SQL should not be displayed at all. Could you please take a look?

dkaukov:
This is expected SA818 behavior. After TX ends, the module keeps SQ closed for a short time before opening it again. You can see the same thing on the status LED.

So I think the app is correctly showing the actual state from the radio module, even when squelch is disabled. To me, hiding this would make the behavior more confusing, not less.

dkaukov:
If we want to hide this transition, I think the cleaner fix would be in firmware by masking the SA818 post-TX state. But that feels like a separate change, not part of this UI PR.

VanceVagell:
Merging this for now to make it easier to review subsequent PRs which seem related to this. Hopefully PR 429 and/or 432 will fix the errant "SQL" indication that's part of this PR.

PR #429, opened 2026-06-23
DSP squelch 

### LINKED ISSUE #428: DSP squelch / auto-mute with SA818 SQ=0
The SA818 hardware squelch is pretty slow. For APRS this can be a problem, especially with stations using low TXDelay. In practice, SA818 often needs something like 0.6–1s TXDelay to decode reliably, which is on the high side.

A common workaround is to set SA818 squelch to `0` and reduce volume, so APRS audio is always available to the decoder. We should see if we can automate this better.

Also, SA818 RSSI does not seem reliable enough for software squelch. It can be high with just static/no real carrier, and it can also be low even when a carrier is present. So RSSI alone is probably not a good “channel busy” signal.

Idea:

* Set SA818 `SQ=0`
* Keep raw audio always going to APRS decoder
* Add DSP-based squelch only for muting speaker/app audio and UI busy state
* Do not clean, filter, or gate APRS decoder input

Research phase:

* Capture/log audio samples with `SQ=0` in different cases:

  * no signal / static
  * strong APRS
  * weak APRS
  * voice
  * quiet carrier
* Log simple metrics:

  * total RMS
  * RMS in 300–1000 Hz band
  * RMS in 1000–2500 Hz band
  * RMS in 2500–4000 Hz band
  * band RMS / total RMS ratios
  * maybe zero-crossing rate
* Check if open static can be reliably separated from real signals/carrier
* Do not rule out a simple total RMS gate yet
* Start simple: no FFT unless needed
* Try cheap filters / rolling windows first

Possible implementation later:

* DSP squelch state machine:

  * open fast
  * close slowly
  * use hysteresis
  * use hang time to avoid chattering
* Auto-learn noise floor when muted/closed
* Use it only for audio mute / scan / busy indicator
* APRS decoder should still receive raw audio all the time

Add-on idea:

APRS busy detector could be added separately later. This would not mute audio, but would try to detect APRS activity specifically, for example by watching AFSK tone energy or decoder state. This can be useful for UI/debugging, but it should not be required for the basic DSP squelch work.

dkaukov:
**DSP SQ Architecture**

1. **Band-pass filter**  
   RX PCM is normalized and passed through a 200-500 Hz FIR band-pass filter using esp-dsp.

2. **Deadband crossing detector**  
   The filtered signal is compared against a squelch-level-dependent deadband. Crossings only count when the signal moves from below `-deadband` to above `+deadband`, or vice versa.

3. **Zero-crossing event stream**  
   Each sample produces either:
   - `sampleRate` if a crossing occurred
   - `0` if no crossing occurred

4. **IIR ZCR smoothing**  
   The event stream is smoothed with a one-pole IIR to produce `iirZcr`, an estimated crossings-per-second value.

5. **Soft squelch state**  
   - Opens immediately when `iirZcr < 30`
   - Closes only after `iirZcr >= 30` continuously for 250 ms

6. **Integration**  
   When Soft SQ is enabled, firmware sets the radio module SQ level to `0` and uses the software squelch state instead of the hardware SQ pin.

Fixes: https://github.com/VanceVagell/kv4p-ht/issues/428


android-src/KV4PHT/app/src/main/res/layout/activity_settings.xml
```diff
@@ -398,6 +398,17 @@
                     android:textSize="16dp"
                     android:layout_marginTop="24dp"/>
 
+                <Switch
+                    android:id="@+id/softSquelchSwitch"
+                    android:layout_width="match_parent"
+                    android:layout_height="wrap_content"
+                    android:text="@string/enable_soft_sq"
```
VanceVagell: Why is this new setting necessary? We should just have one squelch implementation for simplicity, unless there's a very compelling reason to offer a choice.

I have no attachment to the old squelch approach, since it seems like this new approach has better overall performance we can just switch over to it.

dkaukov: I’d keep it as a setting for now. The new squelch depends a lot on the ADC/audio frontend, so I’m not sure it will work well on every board yet. Also it doesn’t have tail suppression, and some people may prefer the old behavior because of that.

So yeah, the new one seems better overall, but I’d be careful about forcing it for everyone right away.

VanceVagell: We should avoid adding features or making changes that will lead to an inconsistent experience across kv4p HT devices, or add potentially unclear configuration options where end users may not understand their purpose.

A key value proposition of kv4p HT is that it is easy to build and use, and once we start having a matrix of things that don't work consistently together or require the user to carefully configure their settings for good performance, we lose that.

I'm inclined to close this PR and encourage you to test it thoroughly until you're confident it will work well across all devices, then we can reopen/merge it (without the setting, we certainly do not want that). But I'll hold off on that in case you want to add anything to the discussion first.

dkaukov: Yeah, I get the concern about keeping things simple. But I don’t have all board variants, so I can’t really prove it works well everywhere by myself. That’s also why I’m worried about forcing the new squelch on everyone without an escape hatch.

I thought Advanced Settings was for this kind of thing — options most users shouldn’t need, but can use if something doesn’t work well on their hardware. I think you said something like that before too. ;)

dkaukov: Hi @VanceVagell, what would you prefer here?

I see a few options:

1. Keep it as-is: switch everyone to the new DSP squelch, with a `DeviceState` flag and visible UI switch as an escape hatch.
2. Keep the flag, but remove the UI switch for now.
3. Move the flag out of `DeviceState` and into a board-specific NVS block.
4. Drop the flag completely and switch everyone to the new DSP squelch.

My preference is still option 1, for the reasons above. But if you don’t want the UI setting, option 2 seems like a reasonable compromise.

VanceVagell:
Let's go with #4, if you're pretty confident this is better than what we have right now (and/or you think it will be fixable if we find issues). We'll just hold the next release until it works well, or we can roll it back if there are serious issues.

dkaukov:
Hi @VanceVagell, done. Removed the setting and flag, so DSP squelch is now the only one used.

This also makes #434 irrelevant from the firmware/SQ-pin side. We can close it unless you want to change how TX/SQL is shown in the UI.

VanceVagell:
Thanks for the flag on #434, I'll do that one purely on the Android side. Thanks for the updates to make the firmware better! Exciting stuff.

PR #432, opened 2026-07-05
Standardize S-meter and add RX overload indication

dkaukov:
## Summary

This updates the radio status meter to use a more standard S-meter scale and adds a clear overload indication for very strong RX signals. https://www.giangrandi.org/electronics/radio/smeter.shtml Includes https://github.com/VanceVagell/kv4p-ht/pull/422

## What changed

- Replaces the previous 1-9 S-meter mapping with a calibrated 16-step meter:
  - S1-S9 for normal signal levels
  - yellow over-S9 bars for stronger signals
  - red overload bar when RSSI indicates RX overload
- Adds an RX overload indication for the condition where the radio module starts "clicking" from an excessively strong received signal.
- Updates the main UI to render the expanded meter and show firmware-reported module state such as `TX` and `SQL`.
- Reports TX audio level through the same meter path so firmware-originated TX activity is visible.


android-src/KV4PHT/app/src/main/java/com/vagell/kv4pht/ui/MainActivity.java
VanceVagell, line 977: This does not seem to work anymore after this PR (the red recording circles do not pulse out of the PTT button when audio is heard). This is fine, since the new S-meter based audio indication replaces it, and I agree it's a helpful well-integrated change.

Please remove the now unused code (here and in the layout XML) related to the older recording visualization since this replaces it.

Note that the new S-meter does NOT need to honor the a11y setting "disableAnimations" because it is not an animation in the a11y sense, it is displaying critical live information.

VanceVagell:
This is excellent, thanks for the improvement. I was never happy with the old S-meter, and this is a big upgrade. I left one inline comment/question, and there are a few high level things to get this ready to merge.

Here is a UX spec for the layout of the meter with the additional levels and new colors, please adjust to this:

image

image

Notes:
* The empty state for the S-meter has all of the markers at #FFFCEA 20% (the light gray). The yellow and red ones should remain that gray until they are active, this avoids drawing unecessary attention to them before they are salient.
* The markers are now 4dp wide, they start at 2dp tall and increase in height by 1dp up to 14dp for the overload indicator. There is a 2dp space between each marker.

Let's simplify the PR's current 6 "over S-9" indicators into 3. The reason for this is two-fold:

1. The typical usage scenario is the S-9 scale, we expect most people to be using this most of the time. With 7 "over S-9" marks on the scale, it visually overemphasizes the less common states and competes for attention. Relatedly, it makes it look like normal S-9 values are "very low signal" because it looks like large parts of the scale are being left untapped.
2. The kv4p HT is not a precision instrument, it's not meaningful to offer a high level of granularity on the over-S9 markers, and suggests a level of granularity we don't really have. And for most hams, this high level of specificity on an HT will not be useful.

Thanks again, looking forward to this better scale!

VanceVagell (changes requested):

android-src/KV4PHT/app/src/main/java/com/vagell/kv4pht/ui/MainActivity.java
dkaukov: If you mean removing the old red-circle recording animation, I’d keep that out of this PR. This PR is just about standardising the S-meter behaviour. Removing the TX audio visualisation is more of a separate UI cleanup, so I can do it in a follow-up PR.

dkaukov:
I applied your styling/spacing feedback and tested the 3-yellow-bar version.

After testing, I changed it back to 16 bars. For me this works better. My signals are often above S9, and with only 3 yellow bars the meter clips too early. It’s harder to see if the signal is going up or down.

The extra yellow bars also line up with the normal 10 dB steps above S9, up to the red overload point around -30 dBm. So it’s not meant to be super precise, but it gives a useful indication instead of just clipping.

Visually I think it still feels close to the proposed design. The yellow bars feel like the normal meter continuing, and the red bar is still the actual warning/overload state.

The switch is in a separate commit, so you can try both versions.

dkaukov:
image

VanceVagell:
Re: "with only 3 yellow bars the meter clips too early"

To clarify, my intent was that each of the 3 bars would cover the same range of values as two of the former bars. 

For example, if one of the old bars covered 10dB above S9, then one of the new bars would cover 20dB above S9. That way you still have the full extent of the scale the radio module can measure via RSSI, but we don't over-emphasize the above-S9 values. These will be most common in bench testing scenarios, whereas field usage should typically be within S9 or just a little above (e.g. people using the kv4p HT in close proximity like a walkie talkie).

Based on what you wrote, I'm assuming the intermediate change you made reduced the extent of the scale, which wasn't my intent. But please let me know if I'm wrong here.

VanceVagell:
Also, there's a small alignment issue where the text indicators are too high, they should be bottom aligned with the scale (see distance from the line I added):

image

dkaukov:
Hi @VanceVagell,  Yes, the 3-yellow-bar version I tried initially reduced the range. Today I also tried the other version, where the full range is kept but each yellow bar is 20 dB instead of 10 dB. It works, but I didn’t like that version much either. It feels too coarse, and it no longer maps nicely to the usual S9+10, S9+20, etc. So I still prefer the 16-bar version with 10 dB steps above S9.

To try the 20 dB version, it’s basically just this change:

```java
int over = (int) Math.floor((dbm - S9_DBM) / 20.0f);
```

For the label alignment: the `TextView` is already bottom-aligned. The gap is mostly from font spacing for letters like `g`, `p`, and `y`. I can push it lower for English, but then some languages, like Chinese, may end up too low / below the S-meter.

VanceVagell:
Please implement the UX spec as provided for the bars, and I'll approve and merge this PR.

Don't worry about the text alignment for now, I may address that in a separate commit.

In terms of whether more or less bars are useful and/or preferred, that would require a usability study with a small number of kv4p HT users to answer definitively. Without that, it's your preference versus my professional judgment of what will ensure the UI is approachable while still having enough information for most people. I believe the current UX design strikes that balance, but would be happy to adjust it in the future if/when I can conduct a study.

I would very much like to conduct a usability study one day for kv4p HT, but it's a significant time investment. I've started a tracking issue for that (see #442), and include this question in a running list we'd like to answer.

dkaukov:
Hi @VanceVagell, done — updated the S-meter to the UX spec: 9 normal bars, 3 over-S9 bars, and 1 overload bar. Also updated the mapping and tests.

PR #436, opened 2026-07-10
Add kv4p_rx_streamer: standalone RX-only WiFi audio streaming firmware

ttimpe:
New firmware alongside the stock sketch that repurposes the kv4p-ht board as an RX-only receiver:

- Live audio as HTTP endless-WAV (PCM16 mono 16 kHz) on :8000/stream.wav, playable directly in VLC and consumable by a decoder server via HTTP GET
- Web UI on port 80: channel table (number, name, frequency, bandwidth, voice/data), tuning, volume/squelch, WiFi setup, browser OTA updates
- Default channels: German Freenet (VHF) or PMR446 (UHF)
- SA818 filters always fully bypassed for a flat audio path (data channels carry VDV-FFSK/NEMO modem signals)
- WiFi STA with AP-fallback provisioning portal, mDNS kv4p-rx.local
- No Bluetooth, no KISS protocol, no TX; radio is permanently in RX

Reuses upstream pin defaults, PCB-revision detection (straps + NVS hwconfig), and the I2S-ADC capture approach from the stock firmware.


dkaukov (changes requested):
Hi @ttimpe, it looks like this PR may have been opened against the wrong branch by mistake. If so, could you please close it? If not, could you remove the commits that delete half of the project and explain the intended use case?

ttimpe:
Yeah totally misclicked.

PR #441, opened 2026-07-16
Fix foreground microphone service access

### LINKED ISSUE #440: Transmit audio is silent when app is backgrounded or screen is off
## Description

Transmit audio works while the Android app is in the foreground, but becomes silent when the app is moved to the background or the screen is turned off. The radio still appears to transmit; only the microphone audio is missing.

This may be related to Android foreground-service microphone permissions or the service not declaring microphone use while the app is backgrounded.

## Steps to reproduce

1. Connect the Android app to a UHF 2.0e board.
2. Transmit while the app is visible and confirm that audio is present.
3. Move the app to the background or turn off the screen.
4. Transmit again.

## Expected behavior

Microphone audio should continue to be sent during transmission when the app is backgrounded or the screen is off.

## Actual behavior

The radio transmits, but no microphone audio is heard unless the app is in the foreground.

## Hardware

- UHF 2.0e board from newdiytech.com

Originally reported by Abraham (M7YAB).

dkaukov:
Fixes: https://github.com/VanceVagell/kv4p-ht/issues/440

Cherry-picked from: https://github.com/VanceVagell/kv4p-ht/pull/393

### DIFF
--- a/android-src/KV4PHT/app/src/main/AndroidManifest.xml
+++ b/android-src/KV4PHT/app/src/main/AndroidManifest.xml
@@ -18,6 +18,7 @@
     <uses-permission android:name="android.permission.FOREGROUND_SERVICE_MEDIA_PLAYBACK" />
     <uses-permission android:name="android.permission.WAKE_LOCK" />
     <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />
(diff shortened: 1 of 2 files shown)


VanceVagell:
Thanks!

PR #443, opened 2026-07-18
Fix soft squelch transition after TX

### LINKED ISSUE #434: [Android] "SQL" should not be displayed briefly after TX when user has set squelch to 0
See PR #422 for details of what's causing this issue, and a description of how this should work.

dkaukov:
Use the radio module SQ pin to pause soft squelch DSP during RX/TX transitions.

While the SQ pin is closed, the module is outputting artificial silence, so the firmware keeps the current soft squelch state instead of feeding that silence into the detector. Once the SQ pin opens, detection continues from the same state.

Fixes #434


VanceVagell:
Thank you!

Issue #427, opened 2026-06-15
Add proper device/connection management UI


dkaukov:
Current snackbar notifications don’t work well once we have USB, BLE, and maybe more than one device.

We need a better way to manage connections.

Things to figure out:

1. Show what device is currently connected

   * USB or BLE
   * device name/id
   * firmware version if available

2. Allow disconnecting

   * User should be able to manually disconnect from the current device.

3. Show available devices

   * USB devices
   * BLE devices
   * Let user pick which one to connect to.

4. Firmware mismatch with BLE

   * BLE devices can’t be flashed over the radio connection.
   * We need to decide what the app should do:

     * block connection?
     * allow limited use?
     * show warning and tell user to update firmware another way?

5. Multiple USB devices

   * Do we support more than one USB radio?
   * Or do we keep it simple and allow only one active device?

Maybe this should be a separate “Devices” screen/tab/pop-up?

VanceVagell:
Just noting (mostly for myself) that this is a high priority issue that must be resolved before the next release. I'm running into the issue where anytime one of my kv4p HT radios powers up (e.g. I'm flashing it) it starts connecting to various Android devices I have via Bluetooth. I even had one race condition where I was plugging it in via USB on one phone, but it connected via Bluetooth to another phone before that first phone negotiated the USB connection!

I want to keep this as lightweight as possible, but also avoid scenarios like I describe above.

It will be something along the lines of a "Pair" option in the Android app that lists currently-broadcasting BLE kv4p HT devices, lets you pair with it, and choose short name for the radio at that point which we save to the radio's NVS via a firmware command. The app will list paired devices, with the option to unpair each one. Whenever a paired device is detected and no other device is connected, it will connect to the paired device. It will never connect to more than one device.

Since most users will not pair via Bluetooth (they will use USB), this is likely to be in settings rather than a top level "..." menu item. However, since the device management list is non-trivial we might need a simple "Paired devices" screen/activity, so we can have the UI controls we need (e.g. pair button, list of devices, unpair buttons) without being super crowded in settings.

It will be rare for people to have >1 kv4p HT, only power users and project developers are likely to be in this situation. So it's less important that we let them name devices that are only ever connected via USB.

(Mostly thinking out loud with this, so I don't lose my thoughts when I put together the design for this issue.)

PR #446, opened 2026-07-22
Implement KISS BLE support in the Android app without a connection manager

dkaukov:
## Summary

Adds automatic BLE support to the Android app for kv4p HT devices using the KISS protocol.

## User-visible behaviour

* When the app starts, it first looks for a connected USB radio.
* If no USB radio is available, it automatically scans for a nearby kv4p HT advertising the BLE KISS service and connects to it.
* No new connection screen or device picker is added. Connection and reconnection remain automatic.
* Radio control, audio, and APRS use the same KISS protocol over either USB or BLE.
* If the connection is lost, the app returns to the existing disconnected state and retries automatically.
* Firmware flashing remains USB-only. The flash action is disabled while connected over BLE.
* On Android 12 and newer, the app requests Nearby Devices permission to discover and connect to the radio.

## Current limitations

* The app connects to the first matching BLE KISS device it discovers.
* Manual device selection and explicit connect/disconnect controls are outside the scope of this PR.


f4dvk:
Hello,

@VanceVagell When will this pull request be available?

Thanks,
Stephane.

VanceVagell:
This is the next one I plan to review, but it will require manual testing
so I need to reflash a device and test it properly. If it's blocking
something you're working on, I can prioritize it.

VanceVagell:
This works really well! I was impressed by how seamless it was to connect.

I'll add a few UX designs to #427 so we can let people pair more concretely to specific devices, which will be required for a bunch of scenarios, but most egregiously ham fests where we don't want everyone's app connecting to random people's radios :) I think we can still keep it simple while making it more deterministic.

PR #449, opened 2026-08-06
TX timeout

dkaukov:
Add a firmware-side TX audio watchdog that releases PTT when host audio frames stop arriving. This prevents the radio from remaining keyed after a USB, Bluetooth, or BLE connection disappears without sending PTT-up.


VanceVagell:
Good QoL fix, thanks!

PR #456, opened 2026-08-18
Implement software RX CTCSS detection and tune soft squelch

### LINKED ISSUE #444: [firmware] Restore RX CTCSS with soft squelch
When we switched the SA818 to `SQ=0` and moved squelch into firmware, RX CTCSS stopped working. `ctcss_rx` is still sent to the module, but the SA818 ignores it for RX audio in monitor mode.

We should add a CTCSS detector in firmware and combine it with soft squelch.

### First, check the audio

Before adding the detector, confirm that CTCSS tones are still present in the SA818 audio output.

Test a few tones across the range with:

* all filters off
* high-pass on
* pre/de-emphasis on

If the SA818 filters the tones out before they reach the ESP32, software detection may not work with the current audio path.

dkaukov:
## Summary

This PR moves RX CTCSS detection from the SA818 module into the firmware DSP pipeline.

The SA818 is now configured with RX CTCSS disabled, while TX CTCSS remains module-controlled. Received audio is analyzed in firmware and gated by `SoftSquelchEffect` when the configured CTCSS tone is detected.

## Changes

- Add DSP detection for all 38 SA818 CTCSS tones.
- Preserve tone index `0` as CTCSS disabled/bypassed.
- Use the table-based `afsk::dsp::DdsOsc` for efficient sine/cosine correlation.
- Analyze audio over a 400 ms window to distinguish adjacent CTCSS tones.
- Combine CTCSS detection with the existing soft-squelch decision.
- Configure the SA818 with RX CTCSS disabled.
- Keep TX CTCSS behavior unchanged.
- Move soft-squelch noise analysis to the 3.5–4.0 kHz portion of the baseband, away from CTCSS and most voice energy.
- Tune the soft-squelch FIR, ZCR threshold, and sensitivity scale for the new analysis band.

Fixes: https://github.com/VanceVagell/kv4p-ht/issues/444

### DIFF
--- a/microcontroller-src/kv4p_ht_esp32_wroom_32/dsp/audioResampler.h
+++ b/microcontroller-src/kv4p_ht_esp32_wroom_32/dsp/audioResampler.h
@@ -115,10 +115,12 @@ public:
   bool begin() {
     captureSamples = 0;
     hasPendingByte = false;
-    return decimator.begin();
+    return decimator.begin(highStop, lowStop);
   }
 
-  bool setFilters(bool highStop, bool lowStop) {
+  bool setFilters(bool useHighStop, bool useLowStop) {
+    highStop = useHighStop;
+    lowStop = useLowStop;
     return decimator.begin(highStop, lowStop);
   }
 
@@ -149,6 +151,8 @@ private:
   size_t captureSamples = 0;
   uint8_t pendingByte = 0;
   bool hasPendingByte = false;
+  bool highStop = false;
+  bool lowStop = false;
 
   size_t decimateFrame(uint8_t *dst) {
     int16_t *pcmWire = (int16_t *)dst;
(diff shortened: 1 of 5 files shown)


VanceVagell:
This seems complex, is the firmware squelch worth the extra downstream complexity it introduces? I'm OK to merge this, just want to spot check that the firmware squelch is adding real benefit.

dkaukov:
I think it's worth it mostly because it keeps things simple for users. They don't need to know that SQ and filters should be disabled for good APRS performance. That's especially important for people using it as a KISS TNC/modem — it should just work.

We discussed this on the original DSP SQ PR and decided to switch everyone to DSP squelch rather than add another setting.

CTCSS is basically a follow-up from that change — RX CTCSS stopped working with `SQ=0`, and you asked me to fix that regression before the next release.

So yeah, firmware gets a bit more complex, but the user side gets simpler.

VanceVagell:
Sounds good, thanks for clarifying the purpose.

PR #457, opened 2026-08-27
Add FreeDV 2400B digital voice mode

dkaukov:
## Summary

Adds FreeDV 2400B digital voice support across the Android app and ESP32 firmware.

Users can switch between analog voice and FreeDV 2400B by long-pressing the Voice navigation button. The caption changes between **Voice** and **2400B** so the active mode remains clear, including after reconnecting the radio.

## User experience

- Long-press **Voice** to toggle analog/FreeDV operation.
- Updates the navigation caption to reflect the radio’s actual mode.
- Restores the correct displayed mode after USB or BLE reconnection.
- Uses the existing squelch control for both analog and FreeDV reception.
- Reports FreeDV squelch state consistently to Android and the RGB LED:
  - solid green while open
  - breathing blue while closed
- Drives the existing S-meter from decoder SNR in FreeDV mode.
- Maps the observed FreeDV quality range from S0 through S9/S9+10.
- Preserves the existing analog voice controls and protocol behavior. The shared RX sampling path now includes ADC clock compensation and a transparent, continuously running FreeDV detector.

## Implementation

- Integrates `esp32-freedv2400b` in the ESP32 firmware for 2400B modulation and demodulation.
- Integrates Codec2 on Android for microphone encoding and received-voice decoding.
- Adds seven-byte digital voice packets to the KV4P protocol.
- Advertises FreeDV support through firmware feature negotiation.
- Adds FreeDV squelch hysteresis, timeout handling, and decoder-quality filtering.


VanceVagell:
This is super cool! I was hoping we could add FreeDV one day, since it works at much lower baud rates than the other digital modes. Just pinging here so you know this is on my radar. I want to test this out, and there will definitely be some UX changes on this one (i.e. almost certainly moving the analog/digital choice into Settings instead of long press, but I want to play around with it to understand it better before I weigh in). Thanks for the amazing work!

dkaukov:
Cool, looking forward to your feedback once you’ve had a chance to try it. The long-press is definitely just a temporary POC UX. I’m especially curious how it behaves in real use. On the bench it keeps working at pretty low SNR, where normal voice is barely understandable.

VanceVagell:
This PR won't build for me in Android Studio (on Windows 11).

What are you using to build it? Are there environment modifications I need to make locally for this to work? It seems to be some C compilation issue around codec2 codebook generation.

dkaukov:
> This PR won't build for me in Android Studio (on Windows 11).

Yeah, Codec2 needs host build tools to generate the codebooks, and Android NDK doesn’t provide them. The current build looks for `cc`, `clang`, or `gcc`.

For now you can probably work around it by installing MSYS2/MinGW or LLVM and putting `gcc`/`clang` on PATH.

For testing, the PR artifact already has the built APK and firmware. I’ll look at making the Codec2 build less painful on Windows.

dkaukov:
@VanceVagell This should no longer be an issue. I moved Codec2 into a separate prebuilt Android library, so the app no longer needs host build tools to compile it.

VanceVagell:
OK I've tested this out, it's very cool. I'm impressed you were able to squeeze decent packet throughput out of this! In much of packet radio, there's huge latency between packets that makes this sort of thing impossible.

Of course the audio quality is only so-so, but it is certainly intelligible.

In terms of how to properly switch between voice modes, please:
1. Remove the long-press on the voice mode tab
2. Add a new setting under "Radio" in settings as such:

Radio <-- existing section
Power [High] <-- existing setting
Voice mode [Analog] <-- new setting

The voice mode drop down has two options: "Analog", "FreeDV 2400". Defaults to "Analog".

I'm not sure of the distinction between "2400" and "2400B", if the "B" is not important lets omit it because space is tight, otherwise you can include that in the menu item label.

I'm not super familiar with what data FreeDV sends, though I've played around with it a bit in the past. Can we send and display the user's callsign on the receiving radio? Not in this PR, but if it's possible it would be a nice value add for using digital over FM to see an indication of who's talking.

dkaukov:
Hi @VanceVagell, updated the UI as suggested — removed the long-press and added **Voice mode** under **Settings → Radio**, with **Analog** and **FreeDV 2400B** options.

I kept the `B` since 2400A and 2400B are different FreeDV modes.

Callsign display sounds useful too. I’d probably keep that as a separate follow-up once this is merged and stable, since it requires changes to the 2400B modem library as well.

PR #459, opened 2026-09-05
Update Android target API to 36

### LINKED ISSUE #458: [Android] High priority: update Android API level
The Play store is now requiring API level 36+, we are on 35+. We will not be able to push a new version of the Android app to the store until this is fixed. Here's the detail:

App must target Android 16 (API level 36) or higher
To provide users with a safe and secure experience, Google Play requires all apps to meet target API level requirements.
From Aug 30, 2026, if your target API level is not within 1 year of the latest Android release, you won't be able to update your app.

How to fix
Your highest non-compliant target API level is Android 15 (API level 35).

To ensure that you can continue to make updates to your app you must:

Update your app to target Android 16 (API level 36) or higher.
Publish a new version of your app to production. Before you do this, you can test your app using internal, closed, or open testing.

dkaukov:
And dependencies upgrades.

Fixes: https://github.com/VanceVagell/kv4p-ht/issues/458

### DIFF
--- a/android-src/KV4PHT/app/build.gradle
+++ b/android-src/KV4PHT/app/build.gradle
@@ -44,32 +44,32 @@ android {
 
 dependencies {
     implementation project(':usbSerialForAndroid')
-    implementation 'androidx.activity:activity:1.10.1'
-    implementation 'org.junit.jupiter:junit-jupiter:5.8.1'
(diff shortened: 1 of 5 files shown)


VanceVagell:
Thank you!

PR #454, opened 2026-08-11
Update firmware README.md

theojalba:
This PR updates the instructions for compiling the firmware:

1. Re-arranged the build options so PlatformIO is first, and marked as preferred. While Arduino IDE is generally a simpler IDE, for a project of this size I believe Arduino IDE is harder to use. The initial configuration of the options and library dependencies is very manual and surfaces a lot of details.
2. Moved the section comparing PlatformIO and Arduino IDE to the top. The user needs that info to make an educated choice about a build environment before delving into the build instructions.
3. Added a few clarifications to the PlatformIO instructions such as the alien head icon and a note on trusting the project.
4. Update the Arduino IDE instructions to include all the libraries that were missing and where to get each. I matched all the libs and versions in the `platformio.ini` file.

The esp32-afk git LFS problem in PlatformIO appears solved. 👍 

We could potentially remove some sections such as the one on renaming `.ino` to `.cpp` and the `Additional Notes` section. For now I left them in.

One other beginner note that may be useful is how to enable Debug Build and show the Serial Monitor so you can see firmware debug messages. Thoughts?

PR #460, opened 2026-09-06
Add developer clone-and-build outline to the root README #445

### LINKED ISSUE #445: [Other] Improve developer instructions on GitHub project main page
Although some of our project sub sections have developer instructions to help people get started, our main project page (e.g. the readme.md in the root) should have a high level outline of how to clone and build the project for yourself.

For example, we've had a couple people recently build only the Android app without building the firmware, not realizing that when building as a developer you must do both because the version mismatch check isn't active until a public release is launched.

AzazelSensei:
The root README only pointed at kv4p.com. I added a short clone-and-build outline so people compiling from git know they need matching firmware and the Android app.

The firmware Arduino library list was already updated in #454, so I left microcontroller-src/README.md alone.

Fixes #445

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,41 @@
 # kv4p-ht
+
 Open source handheld ham radio project kv4p HT
 
 Please see the main project site: https://kv4p.com
+
+## Building from source
+
+The site above is for installing a public release. Compiling this repo yourself means building **both** the ESP32 firmware and the Android app from the same git revision.
+
+The version mismatch check is not active until a public release. If you only rebuild the Android app and leave old firmware on the radio, they will still connect. Flash matching firmware whenever you change firmware or the protocol.
+
+### Clone
+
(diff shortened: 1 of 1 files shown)


dkaukov:
Hi @AzazelSensei, looks good. I think we should just clarify one thing here: the firmware resource bundled in the Android app is still from the last release and isn't updated till next release.

Also, in dev there is no firmware compatibility check, so it's up to the developer to keep them in sync. Usually that means flashing the firmware and building the Android app from the same commit.

AzazelSensei:
Updated the source-build note. The Android app still bundles last-release firmware until the next cut, and there is no compatibility check in dev, so you have to flash firmware and build the app from the same commit.

VanceVagell:
@dkaukov feel free to merge this one when it LGTY.

PR #466, opened 2026-09-10
Fix audio focus during radio reception

### LINKED ISSUE #465: [Android] App holds audio focus while connected even when squelch is closed
When KV4P is connected to radio, the app seems to keep Android audio focus even when squelch is closed and no RX audio is playing. This causes audio in other apps (WhatsApp/Telegram voice messages, music players, etc.) to pause. It happens while KV4P is in the foreground or background. Closing KV4P by dismissing the notification makes other audio work normally again.

**Expected:** KV4P should only take audio focus when it needs to play RX audio and release it when RX stops.

dkaukov:
## Summary

- Request transient audio focus only while firmware reports an open squelch
- Release audio focus when squelch closes, RX ends, the radio disconnects, or the service stops
- Describe RX audio as speech

This allows music, voice messages, and other media to resume when no radio transmission is being received.

Fixes #465


VanceVagell:
I don't have an easy way to test this, but the change overall LGTM. I'll play around with pairing it to various bluetooth devices (e.g. in my car) and listen for audio issues.

Issue #258, opened 2025-04-15
[Android] Add support to receive APRS packets from APRS-IS services


NaiveTomcat:
Due to the lack of transmitting digipeater in [large] regions, user may send out an ARPS message, which is then relayed through the IS network before it reaches the distant recipient, with the reply message relayed by the IS network but never finding a way to get to the rf again to be received by VHF radios. It would be nice to implement a feature that checks and displays messages whose recipient bears the local callsign from an Internet source (when applicable) and even nicer checks and relay these messages to rf with the recipient being one of the recently heard callsigns (on rf).

The latter function is essentially a transmitting digipeater. The display of messages received from Internet can be alike to those received from rf with an extra sign showing this is not heard from RF, and the digipeater function should be equipped with an on-off switch.

VanceVagell:
This is related to #7

Issue #34, opened 2024-10-19
[Android] Resend APRS Message if ack not received, with exponential backoff


VanceVagell:
Right now, kv4p HT will send an APRS "ack" to other radios when it receives an APRS message addressed to the user's callsign. But if kv4p HT itself receives an "ack" for one of its messages, it does nothing with it.

Instead, we should re-send the attempted Message if we don't get an "ack" after some timeout (maybe 10 seconds with exponential backoff, so 10, 20, 40, etc.), with a limit of max retries (let's start with 5).

It should be smart enough to not try to re-send a Message if the user is actively transmitting, or if it's in the middle of decoding another message.

VanceVagell:
No longer true that it "does nothing" with an ack since #33 was added but we still need retry.

Issue #7, opened 2024-10-16
[Android] Add APRS iGate option


VanceVagell:
Add "APRS iGate" option to Settings, and implement digipeating (so it reports messages it hears online, if an internet connection is available).

If you're implementing this, please ask me for a UX design for the UI details.

VanceVagell:
Here are the relevant pages for iGate APRS-IS integration:

https://www.aprs2.net/
https://www.aprs-is.net/SendOnlyPorts.aspx
https://www.aprs-is.net/connecting.aspx
https://www.aprs-is.net/q.aspx

Note that the last page says "receive-only IGates are discouraged on standard APRS frequencies. Please consider a bidirectional IGate that only gates to RF messages for stations heard directly." This means that clients that can only report packets to APRS-IS, and won't be listening to APRS-IS internet traffic on an ongoing basis to send RF packets to locally-heard users, are discouraged.

They're not prohibited, and I think kv4p HT has a decent use case where receive-only iGating might be preferable, since an always-on internet link from the app could have battery implications (there's a decent amount of iGate traffic in busy areas) and cellular data usage implications. Also, the chance that a kv4p HT is the only client that hears a recipient of an APRS-IS message is fairly low versus WIDE2 iGates (e.g. on mountaintop towers).

Finally, since kv4p HT is intended to work 100% offline (i.e. in airplane mode), this would be a feature that would break in that situation. Whereas digipeating (added in #8) continues to have value because it creates a local RF mesh with no internet connection required.

So I think we can revisit iGating in the future, but I'm removing it from release v2.0.0.0 since it is not an immediate priority.

PR #469, opened 2026-09-11
Centralize APRS messaging and fix Sonar findings

dkaukov:
## Summary

This is still WIP. It started as cleanup of the 200+ Sonar issues, but a lot of the APRS code was mixed into `MainActivity`, so it turned into a bigger APRS refactor.

Main changes:

* Moved APRS handling into a new `AprsController`.
* `RadioAudioService` still handles the Android/radio side: location, notifications, APRS-IS connection and actual RF TX.
* `MainActivity` no longer parses APRS packets or accesses the APRS DB directly.
* Added reliable message retries and ACK/REJ handling.
* Added delivery status in the UI: pending / acknowledged / rejected / failed.
* Improved digipeater handling and duplicate suppression..
* Added RF → APRS-IS iGate support.
* Added optional nearby APRS-IS traffic to the APRS feed.
* Added proper handling for more APRS packet types, including status and station capabilities.
* APRS history is bounded and no longer jumps to the bottom while you are reading older entries.
* Reworked the APRS feed so frequently updated station data can show only the latest entry without losing the underlying history.
* Also includes some smaller Sonar fixes found along the way.

## Database changes

The old `aprs_messages` table mixed what we show to the user with details about the actual packet received over RF.

It is now split into:

* `AprsEvent` — a logical APRS event, for example a position, weather report or message.
* `AprsPacket` — the actual packet received or transmitted over RF/APRS-IS, including frequency, path and raw packet data.
* `AprsFeed` — the materialized view used by the normal APRS UI.

This lets us keep several physical packets for one event. For example, a direct packet and a digipeated copy can be stored separately without showing the same thing twice in the UI. Retries, ACKs and APRS-IS forwarding can also stay associated with the same logical event.

The feed is separate from the event history, so things like position, weather, status and station capabilities can show only the latest update for a station while the full event and packet history is still retained underneath.

This also makes iGate support much cleaner. We can distinguish packets received over RF from packets received from APRS-IS, keep the original RF path/raw AX.25 data, and record the corresponding APRS-IS transmission separately.

## APRS-IS / iGate

The app can now:

* Gate eligible packets received over RF to APRS-IS.
* Apply the usual iGate filtering for `NOGATE`, `RFONLY`, Internet/q paths, queries, etc.
* Handle third-party packets without feeding Internet traffic back into APRS-IS.
* Optionally receive a nearby APRS-IS feed and show it together with RF traffic.
* Keep APRS-IS-only events marked separately until the same event is also heard over RF.

APRS-IS traffic is never retransmitted over RF by this implementation.

Fixes:
* https://github.com/VanceVagell/kv4p-ht/issues/7
* https://github.com/VanceVagell/kv4p-ht/issues/258
* https://github.com/VanceVagell/kv4p-ht/issues/34


android-src/KV4PHT/app/src/main/java/com/vagell/kv4pht/aprs/AprsController.java
VanceVagell, line 1: I'll only comment here to avoid comment spam, but it seems most/all of the new files are missing our standard copyright/license notice, please add.

android-src/KV4PHT/app/src/main/res/layout/activity_settings.xml
```diff
@@ -247,6 +247,83 @@
...
+                    android:layout_marginTop="8dp">
+
+                    <TextView
+                        android:layout_width="wrap_content"
+                        android:layout_height="wrap_content"
+                        android:layout_gravity="center_vertical"
+                        android:layout_weight="1"
+                        android:fontFamily="@font/jost"
+                        android:labelFor="@id/aprsHistoryWindowTextView"
+                        android:text="@string/aprs_history_window"
```
VanceVagell, line 263: What is the purpose of this setting/filter? We already use a RecyclerView to display the APRS history, so it will not slow down at all regardless of how large the message history is in the DB. It's not clear to me what the use case is for limiting the history to a certain length.

That said, if you scroll really far back, it would be nice to offer a circular arrow button on the bottom to jump to the current messages rather than having to scroll forever. If that use case was your intent, the "jump to current" arrow button is the preferred approach.

android-src/KV4PHT/app/src/main/res/values/strings.xml
```diff
@@ -109,6 +109,10 @@
     <string name="freq_145_1750" translatable="false">145.1750</string>
     <string name="freq_145_8250" translatable="false">145.8250</string>
     <string name="aprs_my_position_icon">My position icon</string>
+    <string name="aprs_history_window">History</string>
+    <string name="aprs_message_filter">Messages</string>
```
VanceVagell: Please change label to:

"Message filter"

And values to:

"Show all"
"Only mine"

(Also update the Chinese translations afterwards, to match new semantics.)

VanceVagell (changes requested):
The refactoring in this PR is much appreciated, MainActivity is hugely bloated and ready to be broken up like this, thank you!

Most of this seems reasonable, but I want to understand the DB changes better. What motivated that, and could you please explain what the nature of the DB changes is, and why it was necessary? I'm not sure I follow your initial description.

dkaukov:
Hi @VanceVagell, this is still WIP. It actually started as an attempt to clean up the 200+ Sonar issues, but a lot of them were around APRS handling scattered through `MainActivity`, so I started by extracting that into `AprsController`.

The DB changes came from that refactor, but also from thinking ahead about digipeater and iGate support. The current table mixes user-visible APRS events with low-level packet details. Splitting those into `AprsEvent` and `AprsPacket` makes it possible to keep one logical event while still preserving multiple physical packets, for example direct + digipeated copies, retries, ACKs, etc. I think that will make digipeater/iGate work much cleaner later.

For the time-limited history view: it seems to be a fairly common approach in APRS clients, and we already have tickets around being able to purge old APRS history. But if you don't like having that as a user setting, I'm happy to remove it from the UI and keep the controller support internally for now.

I’ll probably need another 2–3 weeks to finish this properly, clean up the remaining behavior, and make the PR easier to review. I definitely don’t mind early reviews though — they’re useful while this is still taking shape. ;)

Legalsynopsis (approved):

Legalsynopsis (approved):

android-src/KV4PHT/app/src/main/java/com/vagell/kv4pht/aprs/AprsController.java
dkaukov: fixed

android-src/KV4PHT/app/src/main/res/layout/activity_settings.xml
dkaukov: RecyclerView helps with rendering, but we still load the whole query result into memory as `List<AprsEvent>`.

The idea here is mainly to keep that bounded, and also give the user a simple way to look only at recent traffic. Time-based history limits also seem to be pretty common in APRS clients.

I also added a hard cap now, so even `All` won't load an unlimited number of events.

android-src/KV4PHT/app/src/main/res/values/strings.xml
dkaukov: done

